PR #33, opened 2023-01-03
feat(skeleton): Add skeleton loading states

damassi:
Addresses [MOPLAT-565]

This adds a few things to the repo:
- `Skeleton` 
- `SkeletonText`
- `SkeletonBox`
- `constants.ts` where we can include any common shared constants (like `DEFAULT_HIT_SLOP`)
- Updated `BackButton` with `hitSlop` prop and pointed at ^

Screenshot 2023-01-03 at 1 29 11 PM

cc @artsy/mobile-platform 

[MOPLAT-565]: https://artsyproduct.atlassian.net/browse/MOPLAT-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<details>
  <summary>📦 Published PR as canary version: <code>3.5.1--canary.33.3837483446.0</code></summary>
  <br />

  ```bash
  npm install @artsy/palette-mobile@3.5.1--canary.33.3837483446.0
  # or 
  yarn add @artsy/palette-mobile@3.5.1--canary.33.3837483446.0
  ```
</details>


pvinis:
have you check it on dark mode?

package.json
```diff
@@ -3,22 +3,23 @@
   "version": "3.5.0",
   "description": "Artsy's design system for React Native",
   "scripts": {
+    "android": "react-native run-android",
```
pvinis, line 6: i cant say i love this, mostly because say `android` and `ios` are far away, but do the same thing. what do you think? i guess we could name them like `run:android` and `run:ios` and then sort alpha. 🤔 but then the next would be `bundle-install` and `pod-install`, which again.. are far. maybe `install-process:bundle-install` and pod? ugh :(

lib/elements/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return <Animated.View style={style}>{children}</Animated.View>
+}
+
+export const SkeletonText: React.FC<TextProps> = ({ children, ...rest }) => {
+  const color = useColor()
+
+  return (
+    <Flex alignSelf="flex-start">
+      <Text {...rest} bg={color("black10")} color={color("black10")}>
+        {children}
```
pvinis, line 37: is this for the sizing of the skeleton basically? 🤔

pvinis: actually, why do we even need this? maybe we can skip? same for the Box below.

package.json
damassi: Alpha works in the long term for anyone who didn't write the code originally and is thus privy to the pattern. Future dev: "i need to do something regarding android... <scans scripts, looks in the index under "a">. Happy to do `run:ios` and `run:android` tho

lib/elements/Skeleton/Skeleton.tsx
damassi: Yup, this is to size the skeleton! Its needed for force a proper outline of the size

pvinis (approved):

PR #51, opened 2023-02-07
feat: finalizing `SpacingUnit`

pvinis:
We now use numbers for DS values, like `mb={2}` for 20 pixels, and strings with `px` for pixel values, like `mb="17px"`.


damassi (approved):
Looks amazing, tyy

PR #59, opened 2023-02-15
feat: less confusing name for darkmode support

pvinis:
quick rename of the themes

## Release Notes

Some renames:
- `v5light` -> `v3light`
- `v5dark` -> `v3dark`

`v5` was a temporary name, picked to make sure we can try dark mode without messing with v3 (or a potential v4 while we figure stuff out). Now that things are set, we can use `v3` for the non-dark-mode DS, and `v3light` and `v3dark` for v3 with added role name layer (like `background` and `onBackground` etc).

In the future, we will either have a `vX` for a non-dark-mode and `vXlight` and `vXdark`, or if the design team decides to incorporate dark mode, then we will only have a `vXlight` and `vXdark`.

<details>
  <summary>📦 Published PR as canary version: <code>8.7.5--canary.59.4179384877.0</code></summary>
  <br />

  ```bash
  npm install @artsy/palette-mobile@8.7.5--canary.59.4179384877.0
  # or 
  yarn add @artsy/palette-mobile@8.7.5--canary.59.4179384877.0
  ```
</details>

### DIFF
--- a/lib/Theme.tsx
+++ b/lib/Theme.tsx
@@ -1,23 +1,25 @@
 import { useContext } from "react"
 import { isString } from "lodash"
 import { ThemeContext, ThemeProvider } from "styled-components/native"
-import { AllThemesType, ThemeV3Type, ThemeV5Type, THEMES } from "./tokens"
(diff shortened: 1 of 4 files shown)


damassi:
I was wondering about this! thanks for the fix

PR #61, opened 2023-02-16
fix(colors): Fix weird color resolution error

damassi:
This fixes the issue we were seeing with reanimated. The problem turned out to be how we were resolving values within the theme. I made it much more boring and that fixed the problem.

<details>
  <summary>📦 Published PR as canary version: <code>9.0.1--canary.61.4189939454.0</code></summary>
  <br />

  ```bash
  npm install @artsy/palette-mobile@9.0.1--canary.61.4189939454.0
  # or 
  yarn add @artsy/palette-mobile@9.0.1--canary.61.4189939454.0
  ```
</details>


lib/tokens.ts
```diff
@@ -5,12 +5,16 @@
...
 import { SpacingUnit as SpacingUnitV3Numbers } from "@artsy/palette-tokens/dist/themes/v3"
 import {
   TextTreatment as TextTreatmentWithUnits,
   TextVariant as TextVariantV3,
 } from "@artsy/palette-tokens/dist/typography/v3"
+import {
+  convertWebSpacingUnitsToMobile,
+  convertWebTextTreatmentsToMobile,
+} from "./utils/webTokensToMobile"
+import { Neg } from "./utils/types"
```
damassi, line 17: @pvinis - Is all of the information in this `tokens` file necessary? I can't help but wonder if 70% of it could just go away. There is so much here for what is effectively

```tsx
const theme = {
  colors: {
    ...standardColors
    ...lightModeColors
  }
}

const themeLight = theme 

const themeDark = {
  ...theme,
  colors: {
    ...standardColors,
    ...darkModeColors
  }
}
```

No need to worry about it now, but we need to follow up and make this a bit more straightforward before stuff starts getting imported into other apps.

damassi: Couple things we can do:
- Move types to `types.ts`
- The `tokens.ts` file deals strictly with token operations _that are required_. Meaning, transformations related to dark / light mode.
- Eliminate any reference to V3 (except for import paths from `palette-tokens` and direct exports from there that include V3 naming references)
- Eliminate any and all intermediate variables. If something exists on `THEME_V3`, just reference it directly. Everything else is noise

lib/tokens.ts
```diff
@@ -171,94 +139,83 @@ export type ThemeV3WithDarkModeSupportType = {
...
-        onSecondaryHigh: this.v3.colors.white100,
-        onSecondaryMedium: this.v3.colors.black5,
-        onSecondaryLow: this.v3.colors.black5,
-        brand: this.v3.colors.blue100,
-        onBrand: this.v3.colors.white100,
-      },
-    }
-  },
+  v3: THEME,
+  v3light: THEME_LIGHT,
```
damassi, line 219: Do we need this key?

lib/tokens.ts
pvinis: yea, happy to try and simplify this file 👌

lib/tokens.ts
```diff
@@ -171,94 +139,83 @@ export type ThemeV3WithDarkModeSupportType = {
 }
 export type AllThemesType = ThemeV3Type & ThemeV3WithDarkModeSupportType
 
+// These are for styled-system
+export type SpacingUnitsTheme = { space: Record<SpacingUnit, any> }
+export type ColorsTheme = { colors: Record<Color, any> }
+
+const THEME = {
```
pvinis, line 146:
```suggestion
const v3 = {
```
why not just name this with the key directly?

lib/tokens.ts
```diff
@@ -171,94 +139,83 @@ export type ThemeV3WithDarkModeSupportType = {
...
+      italic: "Unica77LL-Italic",
+      medium: "Unica77LL-Medium",
+      mediumItalic: "Unica77LL-MediumItalic",
+    },
+  },
+}
+
+const THEME_LIGHT = THEME
+
+const THEME_DARK = {
```
pvinis, line 186:
```suggestion
const v3dark = {
```
same

lib/tokens.ts
```diff
@@ -171,94 +139,83 @@ export type ThemeV3WithDarkModeSupportType = {
...
+    sans: {
+      regular: "Unica77LL-Regular",
+      italic: "Unica77LL-Italic",
+      medium: "Unica77LL-Medium",
+      mediumItalic: "Unica77LL-MediumItalic",
+    },
+  },
+}
+
+const THEME_LIGHT = THEME
```
pvinis, line 184:
```suggestion
const v3light = THEME
```
this definitely needs the treatment below. we need to add the extra layer on v3light.

lib/tokens.ts
pvinis: I didn't understand the question initially but now I do. yes we need that key. on energy, we only use v3dark and v3light. in both cases we use the role colors, like `primary`. so we def need both light and dark, and both of them need the extra layer.

lib/tokens.ts
```diff
@@ -171,94 +139,83 @@ export type ThemeV3WithDarkModeSupportType = {
...
-      },
-    },
-    textTreatments: fixTextTreatments(textVariantsWithUnits),
-  },
-  get v3light() {
-    return {
-      ...this.v3,
-      colors: {
-        ...this.v3.colors,
-        background: this.v3.colors.white100,
```
pvinis, line 198: notice this

lib/tokens.ts
```diff
@@ -171,94 +139,83 @@ export type ThemeV3WithDarkModeSupportType = {
...
-        onBrand: this.v3.colors.white100,
-      },
-    }
-  },
-  get v3dark() {
-    return {
-      ...this.v3,
-      colors: {
-        ...this.v3.colors,
-        background: this.v3.colors.black100,
```
pvinis, line 226: and this. they are different, and both are needed.

pvinis (changes requested):
if I get to it first, I'll update the PR, if not, small changes required.

also, THANK YOU FOR FIGURING THIS OUT!!

damassi:
Actually, gonna close this and open a new PR. I simplified things library wide.

PR #62, opened 2023-02-16
refactor(palette): Simplify Lib 

damassi:
Goes through and simplifies things in a number of places, from moving components out of atomic design structure patterns into elements to the theme and types. There's a lot here but the notable changes are in tokens.ts and types.ts, which could be stripped down considerably. 

Still need to follow up on some of the theme comments from the other PR and ensure everything works in our other apps, but we should make any changes in here. 

cc @artsy/mobile-platform 

<details>
  <summary>📦 Published PR as canary version: <code>10.0.0--canary.62.4194248360.0</code></summary>
  <br />

  ```bash
  npm install @artsy/palette-mobile@10.0.0--canary.62.4194248360.0
  # or 
  yarn add @artsy/palette-mobile@10.0.0--canary.62.4194248360.0
  ```
</details>


.storybook/storybook.requires.js
```diff
@@ -6,32 +6,31 @@ import {
   addParameters,
   addArgsEnhancer,
   clearDecorators,
-} from "@storybook/react-native";
```
pvinis, line 9: it's better to ignore that file. maybe add to eslintrcignore?
it's autogenerated

lib/Theme.tsx
```diff
@@ -1,33 +1,38 @@
 import { useContext } from "react"
 import { isString } from "lodash"
 import { ThemeContext, ThemeProvider } from "styled-components/native"
-import { AllThemesType, ThemeV3Type, ThemeV3WithDarkModeSupportType, THEMES } from "./tokens"
+import { Themes, ThemeType, ThemeWithDarkModeType, THEMES } from "./tokens"
 
-const figureOutTheme = (
-  theme: keyof typeof THEMES | AllThemesType
```
pvinis, line 7: this used to support something like `theme={{colors: {..}, fonts: {..}}}`, but whatever, we should not need that anyway!

pvinis (approved):
i think we are good here. check my changes too please 🙏

damassi:
Thank you so much for going through and fixing things up @pvinis, really (really) appreciated 👍

PR #72, opened 2023-02-24
feat: moving Screen from energy

pvinis:
[MOPLAT-622]

i added `Screen.md` to explain a bit how this works. I think its not very complex.

- I have not finished the react-navigation cleanup from this branch.
- I have not finished the Tabs folder cleanup wither. the Tabs folder has some wrappers/helpers that we use in energy to make Screen and collapsible-tabs work together well. So we have to pull these in too, along with Screen.

After this PR is done, we should be able to use the same https://github.com/artsy/eigen/blob/main/codemods/remove-spacer.ts on energy this time, and import Screen from palette-mobile canary from here. some type errors will be there, most probably the now required `onBack` for headers (because we don't want react-nav here, but also a back button with no action makes no sense), and probably the Tabs folder helpers too. Remove the Tabs folder and palette folder from energy, fix the errors, and hopefully everything works well.

[MOPLAT-622]: https://artsyproduct.atlassian.net/browse/MOPLAT-622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pvinis:
👀

pvinis:
👀

damassi:
Wanted to tighten up some stuff over in Energy first! (remove jotai, react-nanny, just get it as vanilla as possible). Then once that's all stable move it back over here.

PR #87, opened 2023-04-14
fix: reanimated button crash fix

gkartalis:
## Description

Apparently reanimated doesn't like implicit returns inside the hooks 🤷‍♂️


artsyit:
:rocket: PR was released in [`v11.0.14`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.14) :rocket:

PR #89, opened 2023-04-18
feat(progress-bar): Add `animationDuration` prop

damassi:
This adds a new `animationDuration` prop so that we can control the behavior in consuming apps. A common flow is:

- Start at 0
- Animate to 100
- Instantly go back to 0 (animationDuration=0)
- Animate to 100

So this gives us that ability. 

cc @artsy/mobile-platform


artsyit:
:rocket: PR was released in [`v11.0.16`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.16) :rocket:

PR #91, opened 2023-04-25
chore: update react native version

laurabeth:
Brought an issue to practice-mobile re: a quick way to reference palette-mobile stories. We ended up mobbing on these updates:

- bump react-native to 0.70.8
- refactor broken imports to accurately reflect current project structure

PR #92, opened 2023-04-25
build(deps): followup on rn upgrade

gkartalis:
## Description

Noticed that in https://github.com/artsy/palette-mobile/pull/91 we also changed a script in `package.json`, this pr fixes that and also updates cocoapods as per [react native migration guide](https://react-native-community.github.io/upgrade-helper/?from=0.70.7&to=0.70.8)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "setup:artsy": "./scripts/download-fonts",
     "start:reset-cache": "react-native start --reset-cache",
     "start": "react-native start",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "setup:artsy": "./scripts/download-fonts",
     "start:reset-cache": "react-native start --reset-cache",
     "start": "react-native start",
-    "../../storybook-watcher": "sb-rn-watcher",
```
pvinis, line 27: 😆

PR #93, opened 2023-05-01
chore: upgrade reanimated and missed react native upgrade changes

brainbicycle:
updates reanimated in the palette-mobile app and updates some dependencies missed in the react native upgrade


brainbicycle:
gonna close this one but I do think we still want that hermes change, will open another

PR #94, opened 2023-05-01
feat(old-button): Copy old Eigen button over to palette-mobile

damassi:
This attempts to address an issue we've seen where some form of shared mutable state seems to crash eigen in certain (vague) environments by copying our old button over to here and using that, versus reverting Eigen-wide source changes that pointed at our palette-mobile upgrade. 

Had to add react-spring, as that's what the old button uses, but that's an ok compromise for now.


brainbicycle (approved):
⏪

artsyit:
:rocket: PR was released in [`v11.0.20`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.20) :rocket:

PR #95, opened 2023-05-03
fix(import): fix super relative imports

damassi:
Consuming apps (specifically _jest_, in consuming apps) don't like imports that are unspecific, and will lead to strange import behavior. This fixes that in the newly imported button. 

cc @artsy/mobile-platform


artsyit:
:rocket: PR was released in [`v11.0.21`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.21) :rocket:

PR #97, opened 2023-05-04
fix: no more implicit returns in reanimated hooks

brainbicycle:
We believe that implicit returns are not playing well with babel resulting in our js functions in hooks not being converted to "worklets" which causes crashes when reanimated tries to run them on the wrong thread. 

https://artsy.slack.com/archives/C02NAJ2CGLW/p1683148878257549

props @gkartalis @damassi for sleuthing

tangential: Adds a couple missed react native upgrade changes, Hermes-engine failing pod install and gem and ruby settings might not be necessary but good to keep in sync. Button story path broke when we brought back old button.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
+ruby '>= 2.6.10'
+
 gem 'cocoapods', '>= 1.11.3'
 gem "fastlane", "~> 2.210"
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -75,7 +75,7 @@ PODS:
     - FlipperKit/FlipperKitNetworkPlugin
   - fmt (6.2.1)
   - glog (0.3.5)
-  - hermes-engine (0.70.7)
+  - hermes-engine (0.70.8)
   - libevent (2.1.12)
   - OpenSSL-Universal (1.1.1100)
   - RCT-Folly (2021.07.22.00):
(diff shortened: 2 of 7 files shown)


ios/Podfile.lock
```diff
@@ -75,7 +75,7 @@ PODS:
     - FlipperKit/FlipperKitNetworkPlugin
   - fmt (6.2.1)
   - glog (0.3.5)
-  - hermes-engine (0.70.7)
+  - hermes-engine (0.70.8)
```
gkartalis, line 78: was this intentional?? 👀

Gemfile
```diff
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
+ruby '>= 2.6.10'
+
```
gkartalis, line 4: 👏

gkartalis (approved):
thank you!

damassi:
👍

artsyit:
:rocket: PR was released in [`v11.0.23`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.23) :rocket:

ios/Podfile.lock
brainbicycle: yeah believe this should be in lock step with our react native version: https://github.com/artsy/palette-mobile/blob/3c8432e8cb183609b55415c3a0fdc81751aab11e/package.json#LL110C29-L110C29 and we missed it when upgrading, pod install fails without it in the dev app

PR #100, opened 2023-05-05
fix: mark all reanimated hooks functions as worklets explicitly

brainbicycle:
We are continuing to see crashes + red screens due to these functions, in theory this should not be necessary: https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/worklets/#using-hooks
But something seems to be broken with our setup. While we investigate it should be safe to mark these explicitly as workouts since that is what they should be anyway.

Follow-up:
Once we figure out what is broken and fix it or upgrade to 3 we should be able to get rid of this


artsyit:
:rocket: PR was released in [`v11.0.26`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.26) :rocket:

PR #101, opened 2023-05-12
feat: add right button support to entity header

MounirDhahri:
This PR adds support of custom right buttons to the `EntityHeader`. After this PR, we should no longer use the FollowButton and instead use the custom button in the future. It's more scalable and we can use different styles

Screenshot 2023-05-12 at 11 49 17

### DIFF
--- a/src/elements/EntityHeader/EntityHeader.tsx
+++ b/src/elements/EntityHeader/EntityHeader.tsx
@@ -1,36 +1,40 @@
+import { isValidElement, useMemo } from "react"
 import { bullet } from "../../utils/text"
-import { Avatar } from "../Avatar"
+import { Avatar, AvatarSize } from "../Avatar"
 import { Flex, FlexProps } from "../Flex"
 import { Text } from "../Text"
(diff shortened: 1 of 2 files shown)


src/elements/EntityHeader/EntityHeader.tsx
```diff
@@ -1,36 +1,40 @@
...
   initials,
-  name,
   meta,
-  FollowButton,
+  name,
+  smallVariant = false,
   ...restProps
 }: EntityHeaderProps) => {
-  const followButton = FollowButton && (
+  const rightButton = (RightButton || FollowButton) && (
```
MounirDhahri, line 30: prioritize the right button (should overwrite the FollowButton)

dariakoko (approved):
LGTM
Thank you Mounir!

artsyit:
:rocket: PR was released in [`v11.0.27`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.27) :rocket:

PR #102, opened 2023-05-13
chore(screen): make stuff consistent

damassi:
Needing to do some wrangling to make this work between eigen and energy 😓


artsyit:
:rocket: PR was released in [`v11.0.28`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.28) :rocket:

PR #103, opened 2023-05-13
fix(screen): fix safe area

damassi:
Oops! Need the Screen safe area for when stuff scrolls up. Of course of course.... 

Also perfectly centers the center area in the header.


artsyit:
:rocket: PR was released in [`v11.0.29`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.29) :rocket:

PR #106, opened 2023-06-02
fix(header): Make expand

damassi:
When we removed the empty flex, there was nothing left to expand the center text area across the way for the left and right elements.


artsyit:
:rocket: PR was released in [`v11.0.32`](https://github.com/artsy/palette-mobile/releases/tag/v11.0.32) :rocket:

PR #113, opened 2023-06-09
feat (tooltip v2): Add anchorable tooltip

laurabeth:
🧪 👩🏽‍🔬 🧪 👩🏽‍🔬 🧪 👩🏽‍🔬 **MAD SCIENCE: DO NOT MERGE (YET)** 🧪 👩🏽‍🔬 🧪 👩🏽‍🔬 🧪 👩🏽‍🔬 

Just like it says in the header. A tooltip we can anchor to existing components on the screen. 

How it works:
- A `View` wraps the anchor component
- A ref is passed the wrapper `View`, which leverages the timing of the `onLayout` callback to measure the anchor component and its placement within the window (timing is important because there are circumstances in which we could measure and get a set of undefined values)
- The anchor layout + the tooltip dimensions are then used to compute the appropriate X,Y origin for the tooltip
- The origin point is then verified against the window dimensions and adjusted to ensure the tooltip content is visible on the screen

I considered adding our animation library, and left/right positioning, but believe that these should be added in future iterations.


src/elements/ToolTip/v2/geometry.tsx
```diff
@@ -0,0 +1,245 @@
...
+}
+
+const POINTER_SIZE: Size = { height: 12, width: 10 }
+const GUTTER = 5
+
+const computeToolTipTopPlacementOriginPoint = (
+  anchor: Layout,
+  contentSize: Size,
+  padding: Padding | undefined,
+  unconstrained?: boolean
```
damassi, line 72: Here and elsewhere below, we should avoid using unnamed anonymous arguments in function calls and instead use the react-ish props form, with a single interface. That way, arguments can move around and remain more flexible over time, vs everything needing to be locked to a given slot. Additionally, it will cut down on having to redefine all of this stuff for every function up and down this file, with all of the duplication. 

So something like:

```tsx
const computeToolTipTopPlacementOriginPoint = (props: ToolTipProps) => {
  ...
}
```

and then you can thread that same `props` object to each subsequent function call and just grab out what you want/need internally. If additional fields need to be added or removed, a single interface is modified; and if something needs to be narrowed, can `Pick<ToolTipProps, "foo">>` or some such thing.

laurabeth: Thanks for the reminder! I'll make the switch.

src/elements/ToolTip/v2/ToolTip.tsx
```diff
@@ -0,0 +1,224 @@
...
+export type ToolTipPlacementType = "top" | "bottom"
+
+interface ToolTipProps extends PropsWithChildren {
+  title?: string
+  content?: string
+  padding?: Partial<Padding>
+  isVisible?: boolean
+  onClose?: Function
+  placement?: ToolTipPlacementType
+  useInteractionManager?: boolean
```
damassi, line 46: Is this needed? Not seeing it referenced around the PR

src/elements/ToolTip/v2/ToolTip.tsx
```diff
@@ -0,0 +1,224 @@
...
+  height?: number
+}
+
+/**
+ *
+ * @param padding Padding around the content of the tooltip
+ * @param isVisible Whether or not the tooltip is visible
+ * @param onClose Callback for when the tooltip is closed
+ * @param placement Where the tooltip should be placed relative to the child
+ * @param unconstrained Default is false. If this is true, the tooltip will be positioned relative to the window coordinate grid, instead of its immediate parent.
```
damassi, line 60: Can these docs move to the interface up above? That's how we get intellisense docs in typescript; where down here nothing will be picked up

dzucconi:
Cool!

Some screen captures I think would go a long way here?

It's unclear to me how this responds to both scrolling and interacting with the screen boundaries (flipping)?

src/elements/ToolTip/v2/ToolTip.tsx
```diff
@@ -0,0 +1,224 @@
...
+    windowDimensions,
+    unconstrained,
+    pointerProps,
+    adjustedOrigin,
+    measureAnchorRectangle,
+  ])
+
+  const renderContent = () => {
+    if (origin === null) return null
+    return (
```
damassi, line 165: Can we move the contents of this function into the actual FC return down below, remove this intermediate function, and guard against `origin` directly in the function body of the component itself? This keeps the control flow of the component a bit more vanilla

src/elements/ToolTip/v2/ToolTip.tsx
```diff
@@ -0,0 +1,224 @@
...
+      <Box
+        display={isVisible ? "flex" : "none"}
```
damassi, line 167: Can we do something like 
```tsx
if (!isVisible) {
  return null
}
```
in the main function body vs hiding (while still rendering)?

src/elements/ToolTip/v2/ToolTip.tsx
```diff
@@ -0,0 +1,224 @@
...
+        position="absolute"
+        left={adjustedOrigin?.x ?? origin.x}
+        top={adjustedOrigin?.y ?? origin.y}
+      >
+        <Pointer {...pointerProps}>
+          <Box
+            onLayout={(event: LayoutChangeEvent) => {
```
damassi, line 174: Can we move the contents to a `handleLayoutChange` function up above?

src/elements/ToolTip/v2/geometry.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+export interface GeometryOutputs {
+  pointerProps: { pointerPlacement: PointerPlacementType; mx?: number }
+  toolTipOrigin: Point
+  toolTipPlacement: ToolTipPlacementType
+}
+
+// do we have a const file to pull from for these translations?
+const paddingToPx = (
+  padding?: Padding
```
damassi, line 30: Does this need to be optional? Making this required, it will clean up the `undefined` and `?` requirements in the function body

src/elements/ToolTip/v2/geometry.tsx
```diff
@@ -0,0 +1,213 @@
...
+      case 12:
+        return 120
+      case 0:
+      default:
+        return 0
+    }
+  }
+
+  return {
+    top: translate(padding?.top),
```
damassi, line 53: eg here and below

src/elements/ToolTip/v2/geometry.tsx
```diff
@@ -0,0 +1,213 @@
...
+  anchor: Layout
+  contentSize: Size
+  padding?: Padding
+  unconstrained?: boolean
+}
+
+const computeToolTipTopPlacementOriginPoint = (inputs: OriginGeometryInputs): Point => {
+  const { anchor, contentSize, padding, unconstrained } = inputs
+  const { bottom } = paddingToPx(padding)
+  return {
```
damassi, line 73: this is nitpicky, but can we add a line break above this return to make things more legible?

src/elements/ToolTip/v2/geometry.tsx
```diff
@@ -0,0 +1,213 @@
...
+    x: anchor.x + anchor.width / 2 - contentSize.width / 2,
+    y: unconstrained
+      ? anchor.y - anchor.height / 2
+      : anchor.y - contentSize.height - bottom - POINTER_SIZE.height,
+  }
+}
+const computeToolTipBottomPlacementOriginPoint = (inputs: OriginGeometryInputs) => {
```
damassi, line 80: Ditto here (re linebreaks) and below, around `computeToolTipOriginPoint`. Right now it reads as one long run-on thing

src/elements/ToolTip/v2/geometry.tsx
```diff
@@ -0,0 +1,213 @@
...
+}
+
+interface XAxisGeometryInputs {
+  anchor: Layout
+  contentSize: Size
+  padding?: Padding
+  toolTipPlacement: ToolTipPlacementType
+  toolTipOrigin: Point
+  windowDimensions: Size
+}
```
damassi, line 107: line break

damassi (commented):
Looking good @laurabeth - let me pull it down and give it a run 👀

damassi:
Pulling down the branch and it seems the stories don't want to run -- I wonder whats up. Something local on my machine? 

https://github.com/artsy/palette-mobile/assets/236943/1213330f-2b9c-4bf0-a2aa-f20f6c182718

laurabeth:
@damassi @dzucconi Thanks for eyes. I've read all your comments and will circle back to each as I work through them.

src/elements/ToolTip/v2/ToolTip.tsx
laurabeth: Already removed! Good catch though. Will circle back to this when adding animations, but not necessary now.

src/elements/ToolTip/v2/ToolTip.tsx
laurabeth: ✅

src/elements/ToolTip/v2/ToolTip.tsx
```diff
@@ -36,29 +24,29 @@ export interface Padding {
...
-interface ToolTipProps extends PropsWithChildren {
-  title?: string
+/**
+ * ToolTipV2
+ * @param padding Padding around the content of the tooltip
+ * @param isVisible Whether or not the tooltip is visible
+ * @param onClose Callback for when the tooltip is closed
+ * @param placement Where the tooltip should be placed relative to the child
+ * @param unconstrained Default is false. If this is true, the tooltip will be positioned relative to the window coordinate grid, instead of its immediate parent.
+ */
```
damassi, line 34: Sorry! what i meant is you have to annotate the interface itself, in a very specific form. So like:

Screenshot 2023-06-14 at 1 43 00 PM

Screenshot 2023-06-14 at 1 42 15 PM

src/elements/ToolTip/v2/ToolTip.tsx
```diff
@@ -118,8 +105,18 @@ export const ToolTip: FC<ToolTipProps> = ({
     [contentSize, paddingAggregate, unconstrained]
   )
 
+  const measureTooltipContentRectangle = useCallback(() => {
+    if (contentRef.current?.measure) {
+      contentRef.current.measure((x, y, width, height) => {
+        if (isEqual(contentSize, { width, height })) return
+        console.log("contentRef.current.measure", { width, height })
```
damassi, line 112: stray

src/elements/ToolTip/v2/ToolTip.tsx
laurabeth: Done

src/elements/ToolTip/v2/ToolTip.tsx
laurabeth: The not rendering has been problematic, as you know. But, switched. It's broken either way. 😆 😢

laurabeth:
@damassi changes pushed. 

I think the issue will be pretty evident as soon as you run the storybook. Opening the story (ToolTip V2: Bottom placement) from the home screen you can see that first render is completely broken. It's my theory that this is because we're initially getting width & height of 0 for the tooltip content. The math seems to work out that way, anyway.

If you sit on the story and either save the storybook file, or reload the app a few times, eventually the tooltip placement is perfect. Clearly that's not good enough, but it at least says to me that the math is correct and the bug is with the data or when we trigger re-renders.

p.s. there's a bug that popped up w/ `placement="top"` tips, but I have a good idea what's wrong and probably how to fix, should we get the rest of this working.

damassi:
Thanks for the breakdown -- will give it a look once I send out the article PR 👀

PR #120, opened 2023-07-07
feat(Image): enable more styling on the Image element and more

araujobarret:
### Description

This PR enables prop drilling into the Image element, respecting FastImage props types. Summary of changes:

- If we pass width and width inside style prop, the prior one has priority over
- enables to use more properties from the native element such as resizeMode

I checked the usage in Eigen and we use only [here](https://github.com/artsy/eigen/blob/a717ab184cb495fbee8212bc4c557d3425bce435/src/app/Scenes/Home/Components/HeroUnitsRail.tsx#L41) and [here](https://github.com/artsy/eigen/blob/a717ab184cb495fbee8212bc4c557d3425bce435/src/app/Scenes/Article/Components/Sections/ArticleSectionImageCollection/ArticleSectionImageCollectionImage.tsx#L21), so these changes are not breaking anything, but I haven't checked Folio so far 🤔 

https://github.com/artsy/palette-mobile/assets/15792853/25782721-ecd9-415b-a7f1-3575f9df63de

### DIFF
--- a/src/elements/Image/Image.tsx
+++ b/src/elements/Image/Image.tsx
@@ -4,6 +4,7 @@ import { PixelRatio } from "react-native"
 import FastImage, { FastImageProps } from "react-native-fast-image"
 import { Easing } from "react-native-reanimated"
 import { createGeminiUrl } from "../../utils/createGeminiUrl"
+import { useColor } from "../../utils/hooks"
 import { useScreenDimensions } from "../../utils/hooks/useScreenDimensions"
 import { Flex } from "../Flex"
 import { Skeleton, SkeletonBox } from "../Skeleton"
@@ -30,13 +31,15 @@ export const Image: React.FC<ImageProps> = ({
   performResize = true,
   src,
   style,
-  ...imageProps
+  resizeMode,
+  ...flexProps
 }) => {
   const [loading, setLoading] = useState(true)
   const dimensions = useImageDimensions({ aspectRatio, width, height })
+  const color = useColor()
 
   let uri = src
-  if (performResize && dimensions) {
+  if (performResize) {
     uri = createGeminiUrl({
       imageURL: src,
       width: PixelRatio.getPixelSizeForLayoutSize(dimensions.width),
@@ -45,7 +48,7 @@ export const Image: React.FC<ImageProps> = ({
   }
 
   return (
-    <Flex position="relative">
+    <Flex position="relative" {...flexProps}>
       {!!loading && (
         <Flex position="absolute" zIndex={1}>
           <Skeleton>
@@ -59,8 +62,8 @@ export const Image: React.FC<ImageProps> = ({
         transition={{ type: "timing", duration: 400, easing: Easing.sin }}
       >
         <FastImage
-          style={[style, dimensions]}
-          {...imageProps}
+          style={[dimensions, style, { backgroundColor: color("black30") }]}
+          resizeMode={resizeMode}
           onLoadStart={() => setLoading(true)}
           onLoadEnd={() => setLoading(false)}
           source={{
@@ -76,10 +79,6 @@ export const Image: React.FC<ImageProps> = ({
 const useImageDimensions = (props: Pick<ImageProps, "aspectRatio" | "width" | "height">) => {
   const screenDimensions = useScreenDimensions()
 
-  if (!props.aspectRatio && !props.height && !props.width) {
-    return null
-  }
-
   const imageWidth = props.width ?? screenDimensions.width
 
   let imageHeight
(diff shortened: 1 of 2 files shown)


ios/Podfile.lock
```diff
@@ -638,7 +638,7 @@ SPEC CHECKSUMS:
   Flipper-RSocket: d9d9ade67cbecf6ac10730304bf5607266dd2541
   FlipperKit: cbdee19bdd4e7f05472a66ce290f1b729ba3cb86
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 3d02b25ca00c2d456734d0bcff864cbc62f6ae1a
+  glog: 04b94705f318337d7ead9e6d17c019bd9b1f6b1b
```
gkartalis, line 641: hmm, lets avoid comiting that one, its a bit frustrating and it pops up once in a while, theoretically you would be able to get rid of it if you do:

```bash
cd ios && bundle exec pod update glog
```

src/elements/Image/Image.tsx
```diff
@@ -42,7 +45,7 @@ export const Image: React.FC<ImageProps> = ({
   }
 
   return (
-    <Flex position="relative" {...flexProps}>
```
gkartalis, line 45: question: should we also keep the `flexprops` around? Will wait for @damassi to approve, don't have much context there and what it might break in the apps (folio,eigen) 🤔

gkartalis:
@damassi for whenever you have time reminder to have a look on this PR

src/elements/Image/Image.tsx
damassi: Its a nice thing to expose the flex props, so that we can pass margins and padding and so on. We should just `omit(flexProps, ['width', 'height'])` before spreading and that would be good.

src/elements/Image/Image.tsx
```diff
@@ -72,6 +76,10 @@ export const Image: React.FC<ImageProps> = ({
 const useImageDimensions = (props: Pick<ImageProps, "aspectRatio" | "width" | "height">) => {
   const screenDimensions = useScreenDimensions()
 
+  if (!props.aspectRatio && !props.height && !props.width) {
```
damassi, line 79: If there's no width passed, it takes the screen width by default. We should leave this fallback

src/elements/Image/Image.tsx
```diff
@@ -72,6 +76,10 @@ export const Image: React.FC<ImageProps> = ({
...
+    return null
```
damassi, line 80: All of this is checked below, so this block shouldn't be needed. 

The logic is: 
- If there's no width, take the screen width
- If there's no height, take the aspect ratio
- If there's no aspect ratio, show console.error. 

With all of those conditions met, the above check isn't necessary.

damassi: Also, can we update line 91 below to read:

```
console.error("[Image] Error: `aspectRatio` is required if `height` is not provided.")
```

That was some copy pasta from Folio

src/elements/Image/Image.tsx
araujobarret: Humm I see, so any suggestions on how to work with % in this case?
I'm asking because it's a huge limiter not being able to use it, and since we do some resizing with Gemini, when we pass % how should we resize the image?

araujobarret: Actually, we can still do the resize and give the `style` prop a higher priority over `dimensions` in the FastImage style declaration, gonna make another commit in a few minutes with the requested changes.

damassi: Not sure how % relates to how the existing fallback code works. Can you elaborate a bit? This comment is mostly around this additional `if` code, which is unnecessary since we check for fallbacks. Also unnecessary as all of the [relevant conditions have been covered](https://github.com/artsy/palette-mobile/pull/120#discussion_r1273862595). 

More generally, In terms of % and react native, not too sure. We don't ever need to pass % since we always know the screen dimensions ahead of time.

araujobarret: I guess was some crazy experiment with props I made, so I tried to define some `width/height` to Gemini and at the same time make the component stick to 100% with `width/height`.
Because of `style={[ style, dimensions ]}` the percentages were overwritten by the defined width/height causing overflow.

araujobarret:
I addressed all the changes and in the styles, I give preference to the `style` props over `dimensions` but we still resize on Gemini; nothing changes in that regard.
Another thing I've added is some background color as we do with `OpaqueImageView` and also in `Force` when we have no URL present.

src/elements/Image/Image.tsx
```diff
@@ -27,10 +30,13 @@ export const Image: React.FC<ImageProps> = ({
   height,
   performResize = true,
   src,
+  style,
+  resizeMode,
```
damassi, line 34: Should we set a default here?

araujobarret: https://github.com/DylanVann/react-native-fast-image#resizemode-enum If we don't provide uses the default of fast-image which is `cover`

artsyit:
:rocket: PR was released in [`v11.2.13`](https://github.com/artsy/palette-mobile/releases/tag/v11.2.13) :rocket:

PR #128, opened 2023-07-26
feat(Pill): Pill element

araujobarret:
### Description

This PR adds the Pill element.
- Used Figma as the source of Truth
- Animations are still limited, only changing opacity when tapped but will add more complex animations to include the `active` state when pressed in a follow-up PR

Note: The video quality of the miniature is terrible, use fullscreen for details

https://github.com/artsy/palette-mobile/assets/15792853/766a4157-9708-4721-a511-27f8e06d0840

### DIFF
--- a/.storybook/storybook.requires.js
+++ b/.storybook/storybook.requires.js
@@ -42,6 +42,7 @@ const getStories = () => {
     require("../src/elements/List/List.stories.tsx"),
     require("../src/elements/MenuItem/MenuItem.stories.tsx"),
     require("../src/elements/Message/Message.stories.tsx"),
+    require("../src/elements/Pill/Pill.stories.tsx"),
     require("../src/elements/ProgressBar/ProgressBar.stories.tsx"),
     require("../src/elements/Screen/Screen.stories.tsx"),
     require("../src/elements/Separator/Separator.stories.tsx"),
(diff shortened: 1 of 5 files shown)


MounirDhahri (commented):
Sorry that you had to write this all from scratch. I think you didn't have to because I believe we already have some of the states built in Eigen and we just forgot to migrate them to palette-mobile [here](https://github.com/artsy/eigen/blob/ed4952b95f54215850a5e9d0bdc76a691c88baf8/src/app/Components/Pill/Pill.tsx)

araujobarret:
My intention is to ignore Eigen completely, it doesn't reflect our DS accurately.

araujobarret:
For example, `PillSize` doesn't exist in our DS, all Pills have a defined size, at least height.
Lots of inconsistencies with DS, in this case, was easier and more on target creating it from scratch.

MounirDhahri:
I am surprised that Eigen doesn't match our DS :( I was under the impression we used the specs from there which comes with some animations. Since you already wrote the component, I think no need to readjust the previous one anymore and we will only need to update the component in Eigen once this ships

MounirDhahri (approved):

araujobarret:
It has some things that match DS, but some don't unfortunately 😞 . I will open a PR to deprecate the existing in Eigen in favor of Palette as soon as this PR gets merged.

damassi (approved):

PR #137, opened 2023-08-03
feat(Popover): add Popover element

araujobarret:
### Description

We have a `Tooltip` element here but it's very complicated to decouple the existing logic and make it composable, especially for the content and all measuring happening inside `Tooltip`.

This PR uses as base the library `react-native-popover-view` as the base element and creates some custom API around it trying to make it as close as possible to the existing palette-web element.

Limitations:
- the library is not able to show multiple popovers at the same time, in order to do that we need to rely on  the prop `onCloseComplete` before making the next Popover alive

PS: Storybook thumbnail didn't work out so nicely with the last story, nothing we can fix here, and no big deal I believe

https://github.com/artsy/palette-mobile/assets/15792853/03cdc5bc-f95c-4a24-aa4b-36e11719bb7d


dzucconi (approved):
Nice and simple; love to see it. Do we have control over that transition? Would make sense to make it less elastic.

araujobarret:
Good point! We do have `animationConfig` that accepts:
image

araujobarret:
@dzucconi using duration 200 and a simple Ease is way better

https://github.com/artsy/palette-mobile/assets/15792853/e1a2eab8-83c3-48c0-9a87-bf2ddcffabe5

dzucconi:
That scale is change intense. Something like a simple opacity fade along with either a subtle translate in the direction of placement or a very small scale change (like 0.9) would be a bit more elegant. (I will remember to bring this up at the design system sync tomorrow.)

damassi:
Yeah i was gonna say as my feedback, no bounce at Artsy 😆 

Let me play with the transition real quick in this PR so we don't have to do so much back and forth! One sec.

damassi:
Updated the animation a bit, though unfortunately the library isn't that configurable (can only control duration and ease style, not in / out or other transformation props like alpha):

https://github.com/artsy/palette-mobile/assets/236943/91c24e3f-6e09-4808-ae4a-45bdceb3a906

src/utils/shadows.ts
```diff
@@ -0,0 +1,16 @@
+import { THEME } from "@artsy/palette-tokens"
+import { StyleSheet } from "react-native"
+
+export const shadows = StyleSheet.create({
```
damassi, line 4: Sharable constants should live in in [constants](https://github.com/artsy/palette-mobile/blob/main/src/constants.ts) and follow the same form as there (no real need to use `Stylesheet.create`, right? A vanilla object is more flexible in different contexts... _I think_). 

Also, is this worth exporting more globally at this time, if we're only applying to this one surface? If so, these items are from the Design Sys figma doc so if anything the values here should [match up to those there](https://github.com/artsy/palette/blob/main/packages/palette/src/helpers/shadow.ts).

```tsx
export const DROP_SHADOW = {
  shadowColor: THEME.colors["black100"],
  shadowOffset: {
    width: 0,
    height: 2,
  },
  shadowOpacity: 0.23,
  shadowRadius: 2.62,
  elevation: 4,
}
```

araujobarret: No issue not exporting it, since it's used only locally by the Popover I don't see any harm living within the element. Done in https://github.com/artsy/palette-mobile/pull/137/commits/c7199158d6dfdcf596b884aacbed08ca33277916

About shadows, gonna poke DS channel because Shadows are a bit different in RN, as you can see Android pops for shadows are just `elevation` and `shadowColor`, while for iOS is `shadowColor`, `shadowOffset`, `shadowOpacity`, and `shadowRadius`. So we can't have 1:1 from DS shadows in this case(here was an attempt of DROP_SHADOW equivalence), there are several websites that generate Shadows for RN online, but they all do the same, make the shadows equal in Android AND iOS.

[Android elevation](https://reactnative.dev/docs/view-style-props#elevation-android)
[Shadow props](https://reactnative.dev/docs/shadow-props)

araujobarret:
> Updated the animation a bit, though unfortunately the library isn't that configurable (can only control duration and ease style, not in / out or other transformation props like alpha):

Thanks for  making it better, I'm not the best with animations but I see you have an Adobe Flash experience maybe 🅰️ 

Re animations customization, I'm gonna reach the original repo to see if they're open to some custom animations, or at least chose which one we want.

damassi:
> open to some custom animations, or at least chose which one we want.

Yah, seems like an easy but super helpful PR 👍

dzucconi:
Yeah, I think if we can't customize the transition we should disable it completely for the time being.

PR #139, opened 2023-08-11
feat: Add xOffset prop to ToolTip

olerichter00:
This PR is related to [ONYX-60] 

### Description

Adding an `xOffset` prop to `ToolTip` in order to position the tooltip more accurately on an element.

[ONYX-60]: https://artsyproduct.atlassian.net/browse/ONYX-60?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,10 +1,12 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": true,
+    "source.organizeImports": false
   },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   // TODO: Update babel plugin to support absolute paths, and set this to
   // `non-relative`.
+
   "typescript.preferences.importModuleSpecifier": "relative"
 }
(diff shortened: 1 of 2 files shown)


araujobarret (approved):
all the offsets related to this component are way too hackish to support the lack of good positioning this element has, we need to deprecate it as soon as possible, but let's move on with this 👍

olerichter00:
> all the offsets related to this component are way too hackish to support the lack of good positioning, we need to deprecate it as soon as possible, but let's move on with this 👍

Yeah, it's pretty hacky and doesn't really work with font scaling anyway. Do you have any ideas on how the positioning could be improved to make the offsets unnecessary?

araujobarret:
It's like game logic, we need a better overflow detection mechanism, so then we could use the absolute positioning with 100% of its capability, touching the Tooltip more I'd avoid it, I worked on #137 but the library is not 100%, the backdrop of the Popover blocks interaction with elements underneath, so we need a different Popover to use it without glitches.

PR #140, opened 2023-08-11
feat: Add onPress to ToolTip

olerichter00:
This PR resolves [ONYX-60] 

### Description

Add `onPress` prop to `ToolTip` to be able to run code when the user presses/dismisses the tooltip.

[ONYX-60]: https://artsyproduct.atlassian.net/browse/ONYX-60?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/elements/ToolTip/ToolTip.tsx
+++ b/src/elements/ToolTip/ToolTip.tsx
@@ -16,6 +16,7 @@ interface ToolTipProps {
   enabled?: Boolean
   initialToolTipText?: string
   maxWidth?: number
+  onPress?: () => void
   position?: "TOP" | "BOTTOM"
   tapToDismiss?: boolean
   testID?: string
@@ -55,6 +56,7 @@ export const ToolTip: React.FC<ToolTipProps> = ({
   enabled = true,
   initialToolTipText,
   maxWidth,
+  onPress,
   position = "TOP",
   tapToDismiss = false,
   testID,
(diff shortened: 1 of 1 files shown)


src/elements/ToolTip/ToolTip.tsx
```diff
@@ -16,6 +16,7 @@ interface ToolTipProps {
   enabled?: Boolean
   initialToolTipText?: string
   maxWidth?: number
+  onPress?: () => void
```
araujobarret, line 19: nit: does `onToolTipPress` also dismiss the Tooltip? If yes then I'd suggest changing to `onDismiss` to be more assertive.

olerichter00: It does, but only if `tapToDismiss` is set to true. So, I thought `onPress` would be more accurate because it's also called when the tooltip doesn't dismiss.

araujobarret: ah gotcha, then it makes sense `onPress` 👍

araujobarret (approved):

artsyit:
:rocket: PR was released in [`v12.0.3`](https://github.com/artsy/palette-mobile/releases/tag/v12.0.3) :rocket:

PR #144, opened 2023-08-30
feat(TabsWithHeader): show BelowTitleHeaderComponent if we don't display the title

araujobarret:
### Description

The check to not display the title was also affecting `BelowTitleHeaderComponent`, if we don't want to display the large `title` but still want to display `BelowTitleHeaderComponent` we couldn't with the current logic.

The changes here allow us to use `showLargeHeaderText={false}` and still render the `BelowTitleHeaderComponent` if we want.

PS: I checked Eigen and it looks like we never used `showLargeHeaderText={false}` before.

https://github.com/artsy/palette-mobile/assets/15792853/b5d8feb0-0b4a-4dda-8db4-2bedd88a6229

### DIFF
--- a/src/elements/Tabs/Tabs.stories.tsx
+++ b/src/elements/Tabs/Tabs.stories.tsx
@@ -74,10 +74,11 @@ storiesOf("Tabs", module)
   ))
   .add("Tabs with header", () => (
     <Tabs.TabsWithHeader
-      title="My header"
+      title="Artist Header"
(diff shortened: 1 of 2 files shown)


araujobarret:
Going to merge it, will address comments/requests later 👍

artsyit:
:rocket: PR was released in [`v13.0.3`](https://github.com/artsy/palette-mobile/releases/tag/v13.0.3) :rocket:

PR #148, opened 2023-09-12
feat(Image): expose gemini resize mode from Image

araujobarret:
### Description

There's some discussion going on around the Image and Gemini params([here](https://artsy.slack.com/archives/C05EEBNEF71/p1694436720499889) and [here](https://artsy.slack.com/archives/C02BAQ5K7/p1690277621906139)), while that's not settled I'm exposing the Gemini resize_to param to fix some issues we're having because of the default `resize_to` to `fit`.

### DIFF
--- a/src/elements/Image/Image.tsx
+++ b/src/elements/Image/Image.tsx
@@ -3,7 +3,7 @@ import { useState } from "react"
 import { PixelRatio } from "react-native"
 import FastImage, { FastImageProps } from "react-native-fast-image"
 import { Easing } from "react-native-reanimated"
-import { createGeminiUrl } from "../../utils/createGeminiUrl"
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`v13.0.7`](https://github.com/artsy/palette-mobile/releases/tag/v13.0.7) :rocket:

damassi:
@araujobarret - when you get back from OOO lets follow up and rename `geminiResizeMode` -> `resizeMode` so we're not encoding service names in our api 🙏

araujobarret:
React native Image has this property already, also FastImage(comes from RN), the closest name was the one I used, but open to suggestions

damassi:
Ah, I see. Aight! Np then 👍

PR #152, opened 2023-09-18
feat: change port to avoid conflict with Eigen

anandaroop:
Changes the default port for palette-mobile's Metro bundler from 8081 to 8082.

This avoids runtime conflicts with Eigen, so both projects can be run simultaneously without any cryptic `address already in use` errors.

https://github.com/artsy/palette-mobile/assets/140521/913bde7b-e669-41f3-8ae4-00c825127867


README.md
```diff
@@ -54,6 +54,8 @@ yarn start
 
 And then either open Xcode and run, or Android Studio, or run `yarn ios` or `yarn android` from the command line, or just open up the simulator if you have done this before.
 
+If you see an error about port 8081 being busy (`address already in use :::8081`), you likely need to quit another running instance of the Metro bundler, such as the one launched via `yarn start` in Eigen.
+
```
gkartalis, line 58: ✨ awesome! 

Wandering though after our conversation if palette should run to a different port to avoid that 🤔 

we could either add it on demand like this:
```sh
yarn start --port 9988
```

Or actually change the start script in `package.json` to have that all the time 🤷

Thoughts?

gkartalis (approved):

README.md
anandaroop: Yes, even better! I took a swing at this in 6b1b943ca4f6f508b6f21e0f9bda9f1004a426d6 but it didn't quite seem to work. 

Requests still seemingly go to the Eigen bundler on :8081, and the simulator is just a blank white screen. Any idea what's up?

gkartalis: oh, curious if removing extra port from the `android` and `iOS` if it works 🤔

anandaroop: It didn't work with or without that I think 🤷🏽

anandaroop: Yeah no rush! Tomorrow is fine.

One more point. If I kill the Eigen bundler from running simultaneously I get this in the sim — 

Screenshot 2023-09-18 at 1 31 32 PM

gkartalis: Okay I think I figured it out.

The `yarn ios` command builds the app and has a hardcoded env var called `export RCT_METRO_PORT=8081` that builds the app to listen to port 8081 by default.

we can add `export RCT_METRO_PORT=8082;` in front of every build script ("android" and "ios") and also add it on the `start` and `start reset cache` scripts and we're good! 

Just tested it out.

Another change that needs to be done is in [project.pbxproj#L264](https://github.com/artsy/palette-mobile/blob/6c1dbf1db4d22dbe37ff77683345aaafaa94ca6c/ios/PaletteMobile.xcodeproj/project.pbxproj#L264) file, more specifically we should change the 8081 from there to 8082.

After the changes above we should be good to go.

anandaroop:
And I removed the doc update, since it is not necessary with the new setup

gkartalis (approved):
🙌🚢🙌🚢🙌🚢🙌🚢🙌🚢🙌🚢

gkartalis (approved):

PR #156, opened 2023-09-28
feat(Pill): adjust horizontal padding of profile variant

araujobarret:
### Description

Small UI correction in the profile variant, horizontal padding should be 1 instead of 2.

image


artsyit:
:rocket: PR was released in [`v13.0.15`](https://github.com/artsy/palette-mobile/releases/tag/v13.0.15) :rocket:

PR #157, opened 2023-09-28
fix: remove unnecessary padding from IconComponent in Message

MounirDhahri:
This PR resolves [] 

### Description

This PR removes the unnecessary padding from IconComponent in Message that gets added even there is no X icon shown.
Group 1 (8)

cc @dblandin


artsyit:
:rocket: PR was released in [`v13.0.14`](https://github.com/artsy/palette-mobile/releases/tag/v13.0.14) :rocket:

PR #163, opened 2023-11-10
Fix(ONYX-508): Fix vertical centering of text within pills

dariakoko:
This PR resolves [ONYX-508] 

### Description
Fixed vertical centering of text within pills

The text within the pill was not centred. In this PR I set the font weight for the text component to have either the default height of a custom left Icon (18px) or the height of the close icon (15px)

This PR changes only the text within the pill for the pills without the close button and pills with the close button (on the screenshot you can see the changes to lines 1, 2, 4, 5)

| Before | After |
| --- | --- |
| simulator_screenshot_1BD37656-E881-499C-8901-DC949E469AD2 | simulator_screenshot_AAD14F12-C54E-4945-B966-A543EC1264DD |

[ONYX-508]: https://artsyproduct.atlassian.net/browse/ONYX-508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/elements/Pill/Pill.tsx
+++ b/src/elements/Pill/Pill.tsx
@@ -68,7 +68,7 @@ export const Pill: React.FC<PillProps> = ({
         )}
         {Icon && <Icon fill={color} ml={-0.5} mr={0.5} />}
 
-        <Text variant="xs" color={color}>
+        <Text variant="xs" color={color} lineHeight={Icon ? "18px" : "15px"}>
           {children}
         </Text>
 
--- a/src/elements/Pill/Pill.tsx
+++ b/src/elements/Pill/Pill.tsx
@@ -7,7 +7,7 @@ import { CloseIcon } from "../../svgs"
 import { IconProps } from "../../svgs/Icon"
 import { Flex, FlexProps } from "../Flex"
 import { Image } from "../Image"
-import { Text, useTextStyleForPalette } from "../Text"
+import { Text } from "../Text"
 
 export const PILL_VARIANT_NAMES = ["default", "search", "filter", "profile", "badge"] as const
 export type PillState = "default" | "selected" | "disabled"
@@ -37,7 +37,6 @@ export const Pill: React.FC<PillProps> = ({
   onPress,
   ...rest
 }) => {
-  const textStyle = useTextStyleForPalette("xs")
   const stateString = selected ? "selected" : disabled ? "disabled" : "default"
   const color = TEXT_COLOR[variant][stateString]
   const showCloseIcon =
@@ -69,7 +68,7 @@ export const Pill: React.FC<PillProps> = ({
         )}
         {Icon && <Icon fill={color} ml={-0.5} mr={0.5} />}
 
-        <Text style={textStyle} color={color}>
+        <Text variant="xs" color={color} style={{ marginBottom: 2 }}>
           {children}
         </Text>
 
--- a/src/elements/Pill/Pill.tsx
+++ b/src/elements/Pill/Pill.tsx
@@ -7,7 +7,7 @@ import { CloseIcon } from "../../svgs"
 import { IconProps } from "../../svgs/Icon"
 import { Flex, FlexProps } from "../Flex"
 import { Image } from "../Image"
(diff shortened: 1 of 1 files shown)


olerichter00 (commented):
Habe you checked if this works with font scaling?

src/elements/Pill/Pill.tsx
```diff
@@ -68,7 +68,7 @@ export const Pill: React.FC<PillProps> = ({
         )}
         {Icon && <Icon fill={color} ml={-0.5} mr={0.5} />}
 
-        <Text variant="xs" color={color}>
+        <Text variant="xs" color={color} lineHeight={Icon ? "18px" : "15px"}>
```
araujobarret, line 71: I don't think we should change the lineHeight of a Text element, that's linked to the variant of it. 
Maybe the right approach would be changing the icon size/positioning?

dariakoko: Do we have a guideline on how the pill should look like? 
When increasing the close icon size to 18 (default value, I believe), it looks huge

anandaroop:
Would also suggest running this by Barney — if I recall correctly his intention was to have the centering account for the vertical distance from the cap height to the descender (see comment [here](https://artsyproduct.atlassian.net/browse/ONYX-508?focusedCommentId=59526)).

src/elements/Pill/Pill.tsx
dblandin: Would [this useTextStyleForPalette helper](https://github.com/artsy/palette-mobile/blob/b48e4ff2722da7c5a0d8062609d3b82debb64efd/src/elements/Text/helpers.ts#L31-L46) be helpful here? From the comment (and [this PR](https://github.com/artsy/eigen/pull/5615)), it looks like it was created with the explicit purpose of centering text within a component.

araujobarret: We have something on [Figma](https://www.figma.com/file/gZNkyqLT8AU3T61tluVJyB/Artsy-3.1-Design-System?type=design&node-id=6726-2087&mode=design&t=9GimAyltawyEPVNV-4) and [palette-web storybook](https://palette-storybook.artsy.net/?path=/story/components-pill--variants).
I think it lies around the position of the Icon, it might be that case when we need just a slightly marginTop -1 but not quite sure.
What @dblandin suggested could be a way, getting the lineHeight of the variant and maybe using to change not the Icon size, but the container around it

dariakoko: @dblandin I don't think this helper would be useful here. For the Pill, we set `variant="xs"` for the Text component. I see `useTextStyleForPalette` helper would work for the buttons

dariakoko: @dblandin the helper is actually useful, but it makes the text's line height 13px which is too small for the pill
simulator_screenshot_FEF82178-4BCB-4750-A3DF-2671935C1179

dariakoko: image

dariakoko: It looks more centred

olerichter00 (approved):

src/elements/Pill/Pill.tsx
araujobarret: with `useTextStyleForPalette` `xs` the returned `lineHeight` is 13?

dariakoko: Yes, 13.3
image

araujobarret: oh boi, it should be 20 😢 
I don't have time today but tomorrow I do, would be happy to pair to check what's going on in more depth, something might be off from our helpers

dariakoko: Let's check it tomorrow, sure
Will put some time into your calendar - feel free to change

dariakoko:
With @araujobarret added a custom margingBottom: 2 to the Text component
Looks good
| iOS | android |
| --- | --- |
| image | image |

PR #167, opened 2023-11-17
fix: patch upgrade for rn ios build issue, fix port not being respected

brainbicycle:
This PR resolves [] 

### Description

- Updates react native to patch version that fixes build issues on iOS on Xcode 15
- Adds a workaround customized port for metro not being respected when running from Xcode

Follow-ups:
- I am not sure having a custom port is worth it, I think the idea is that you can run multiple mobile repos without metro servers interfering but this little bit of customization just sent me down a rabbit hole and requires changing the defaults in a bunch of scripts


gkartalis (approved):
Thanks! 

> I am not sure having a custom port is worth it, I think the idea is that you can run multiple mobile repos without metro servers interfering but this little bit of customization just sent me down a rabbit hole and requires changing the defaults in a bunch of scripts

oof must have been annoying ~ It was meant to help basically in case you need to run both eigen + palette-mobile at the same time - introduced [here](https://github.com/artsy/palette-mobile/pull/152) by me and @anandaroop

anandaroop:
Oof yeah sorry for the rabbitholing. 

What is meant by "changing the defaults in a bunch of scripts" – I don't see that in the diff here, is that over in Eigen?

artsyit:
:rocket: PR was released in [`v13.0.23`](https://github.com/artsy/palette-mobile/releases/tag/v13.0.23) :rocket:

PR #182, opened 2024-02-02
fix(tabs): Ensure tabs use default active opacity

damassi:
### Description

Noticed this was out of sync with the rest of our Touchable opacity configurations, so brought it in line.


artsyit:
:rocket: PR was released in [`v13.0.35`](https://github.com/artsy/palette-mobile/releases/tag/v13.0.35) :rocket:

PR #183, opened 2024-02-02
feat(tabs): Update type to allow title to render jsx

damassi:
### Description

Makes it so that we can render jsx in the tabs title


artsyit:
:rocket: PR was released in [`v13.0.36`](https://github.com/artsy/palette-mobile/releases/tag/v13.0.36) :rocket:

PR #186, opened 2024-02-14
feat: add Screen.LazyScrollContext

MounirDhahri:
### Description

This PR comes as a follow-up on https://github.com/artsy/eigen/pull/9832#discussion_r1485212236

It adds a wrapper to our `Screen` component that exposes some props useful to support lazy loading.

https://github.com/artsy/palette-mobile/assets/11945712/09665f03-a50e-46f3-b267-332f51413c2e

This will be used as follows:
```typescript
const App () => {
  return (
    <Screen.LazyScrollContext<NotificationT> keyExtractor={keyExtractor}>
      {(props) => {
        return (
          <FlatlistComponent
            ...
            renderItem={({ item }) => {
              return <ActivityItem notification={item} isVisible={props.hasSeenItem(item)} />
            }}
            onViewableItemsChanged={props.onViewableItemsChanged}
            viewabilityConfig={props.viewabilityConfig}
          />
        )
      }}
    </Screen.LazyScrollContext>
  )  
}

```


src/elements/Image/Image.tsx
```diff
@@ -24,6 +24,15 @@ export interface ImageProps extends CustomFastImageProps {
...
   geminiResizeMode?: GeminiResizeMode
+  /**
+   *
+   * Delay rendering component
+   * As long as this is set to true, the component will render a skeleton and we won't load the image
+   * This is useful for cases where we want to delay the rendering of the image for performance issues
+   *
+   **/
+  //
+  delayRendering?: boolean
```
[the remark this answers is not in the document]
damassi: showLoadingState

src/elements/Image/Image.tsx
```diff
@@ -35,12 +44,21 @@ export const Image: React.FC<ImageProps> = ({
   style,
   resizeMode,
   geminiResizeMode,
+  delayRendering = false,
```
damassi, line 47: Should this default to `true` since that's how it was? And then in areas where there are performance concerns we toggle it off? Or do you think that its currently risky enough to where we should always avoid the skeleton? (Totally fine with that; just curious because this will lead to an app-wide change in behavior)

src/elements/Screen/LazyScrollContext.tsx
```diff
@@ -0,0 +1,27 @@
+import { ReactElement } from "react"
+import { useHasSeenItem } from "./hooks/useHasSeenItem"
+
+export function LazyScrollContext<T>(props: {
```
damassi, line 4: This isn't exactly a context, so this could be confusing down the line. How about `LazyFlatlist`? 

Also, this seems useful outside of the screen. Can we move it to the `elements` folder? It's (necessary) relationship with the Flatlist (independent of screen) also makes me think that it doesn't belong here. 

Whats yr thinking? (I could be missing something)

damassi (approved):
A couple of minor comments but otherwise looks great 👍. Eager to try this out in Folio.

gkartalis (approved):

src/elements/Image/Image.tsx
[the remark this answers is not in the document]
MounirDhahri: logical

src/elements/Image/Image.tsx
MounirDhahri: I actually don't want to have an app-wide change here and it's why I went with `false` as a default value. By default, I want to keep the existing behavior, unless explicitly specified which is what we will do in the alerts screen.

src/elements/Screen/LazyScrollContext.tsx
MounirDhahri: I am hesitant to name this `LazyFlatlist` because I feel that gives the impression it's a `Flatlist` although it's just the props that should be injected to support the lazy loading behaviour. How about `LazyFlatlistProps` ?

olerichter00 (approved):
Nice 🌟

src/elements/Screen/LazyScrollContext.tsx
damassi: Its a component tho, so props wouldn't make sense (and `ThingProps` is typically our interface name.)

I don't think folks will be confused with just LazyFlatlist, because instantly it will be known that a) it lacks all normal flat list props; and b) its associated with an actual flatlist. Its fairly common linguistically to describe something internal to something as as a operating on b. Ie:

```tsx
<WarmDinner>
  <Dinner>
</WarmDinner>
```

`WarmDinner` warms the dinner, its not _the_ dinner; it acts upon it, like the LazyFlatlist acts upon a Flatlist, by making it lazy.

MounirDhahri: Thanks for explaining. It's clear to me now.

src/elements/Image/Image.tsx
damassi: Oh ok, i was thinking that we currently do this everywhere, and this would flip it off 👍

artsyit:
:rocket: PR was released in [`v13.1.4`](https://github.com/artsy/palette-mobile/releases/tag/v13.1.4) :rocket:

damassi:
@MounirDhahri, one last request -- could you update the PR title and description example with the latest, for future reference?

MounirDhahri:
with pleasure

PR #189, opened 2024-02-21
fix: content jump after switching tabs

MounirDhahri:
This PR resolves [] 

### Description
This PR fixes a jump on the content when switching between tabs on the activity screen. The behaviour we want to support can not be done with Moti and I had to therefore use reanimated instead

**Before**

https://github.com/artsy/palette-mobile/assets/11945712/565ea78a-58a7-4302-b697-7570f82418b4

**After**

https://github.com/artsy/palette-mobile/assets/11945712/079e5d57-c0cb-4b5e-b312-4ade313acc86


olerichter00 (approved):
Looks good 🌟

PR #190, opened 2024-02-21
fix: title centering when no right element is present

MounirDhahri:
### Description

This PR fixes an issue where the title is not centered if there is no right element provided.

For some reason we specify `hideTitle` and `hideRightElements` as props, which doesn't make much sense to me because I think if you don't want to show an element, simply pass nothing. Anyway, here i am bring this to how it was before my previous changes

**Before**
Screenshot 2024-02-21 at 12 46 21

**After**

https://github.com/artsy/palette-mobile/assets/11945712/96453a65-cb3e-4eb7-ad9f-283104c3eb67

### DIFF
--- a/src/elements/Screen/Header.tsx
+++ b/src/elements/Screen/Header.tsx
@@ -52,15 +52,21 @@ export const Header: React.FC<HeaderProps> = ({
       alignItems="center"
       width="100%"
     >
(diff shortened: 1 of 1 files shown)


olerichter00 (approved):
Thank you 🙏

artsyit:
:rocket: PR was released in [`v13.1.8`](https://github.com/artsy/palette-mobile/releases/tag/v13.1.8) :rocket:

PR #192, opened 2024-02-23
feat: Add seperatorComponent prop to StickySubHeader

olerichter00:
### Description

Add seperatorComponent prop to StickySubHeader.


artsyit:
:rocket: PR was released in [`v13.1.10`](https://github.com/artsy/palette-mobile/releases/tag/v13.1.10) :rocket:

MounirDhahri:
As a follow-up on this @olerichter00 , do we intentionally want to use this separator colour only here? I feel it might make more sense to make this the default colour everywhere and avoid having multiple separator colours, What do you think?

PR #195, opened 2024-02-27
fix: broken Image quality on android devices

MounirDhahri:
This PR resolves [] 

### Description

This PR fixes the quality issues we noticed in the image component from FastImage and reported during QA. We decided to use the default Image component from RN on Android while waiting for a fix in FastImage.

**After**
Screenshot_1709055076


gkartalis (commented):
> This PR fixes the quality issues we noticed in the image component from FastImage and reported during QA. We decided to use the default Image component from RN on Android while waiting for a fix in FastImage.

Did we open an issue for fastimage? Is it an existing issue on their repository?

PR #200, opened 2024-03-04
fix: Align right header element correctly

olerichter00:
Resolves https://www.notion.so/artsy/Manage-saves-font-looks-broken-on-the-offer-page-update-font-size-according-to-design-and-center-bu-b344bd5a99f941fea5df5d4048679090?pvs=4

- Related Eigen PR: https://github.com/artsy/eigen/pull/9928

### Description

This pull request corrects the text alignment and centering in the screen header component.

| Before | After |
| --- | --- |
|Simulator Screenshot - iPhone 15 Pro - 2024-03-04 at 16 10 10 | Simulator Screenshot - iPhone 15 Pro - 2024-03-04 at 15 55 22|

<details>
  <summary>📦 Published PR as canary version: <code>13.1.18--canary.200.1559.0</code></summary>
  <br />

  ```bash
  npm install @artsy/palette-mobile@13.1.18--canary.200.1559.0
  # or 
  yarn add @artsy/palette-mobile@13.1.18--canary.200.1559.0
  ```
</details>

### DIFF
--- a/src/elements/Screen/Header.tsx
+++ b/src/elements/Screen/Header.tsx
@@ -7,6 +7,7 @@ import { DEFAULT_HIT_SLOP } from "../../constants"
 import { ArrowLeftIcon } from "../../svgs/ArrowLeftIcon"
 import { useScreenDimensions } from "../../utils/hooks"
 import { Flex, FlexProps } from "../Flex"
+import { Spacer } from "../Spacer"
 import { Text } from "../Text"
 import { Touchable } from "../Touchable"
 
@@ -56,9 +57,8 @@ export const Header: React.FC<HeaderProps> = ({
       justifyContent="space-between"
       alignItems="center"
     >
-      <Center animated={animated} titleProps={titleProps} title={title} hideTitle={hideTitle} />
-
       <Left leftElements={leftElements} onBack={onBack} hideLeftElements={hideLeftElements} />
+      <Center animated={animated} titleProps={titleProps} title={title} hideTitle={hideTitle} />
       <Right rightElements={rightElements} hideRightElements={hideRightElements} />
     </Flex>
   )
@@ -105,6 +105,7 @@ const Center: React.FC<{
   title: HeaderProps["title"]
 }> = ({ animated, hideTitle, titleProps, title }) => {
   const { scrollYOffset = 0, currentScrollY = 0 } = useScreenScrollContext()
+  const { width: screenWidth } = useScreenDimensions()
 
   if (hideTitle) {
     return null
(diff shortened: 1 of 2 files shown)


src/elements/Skeleton/Skeleton.tsx
```diff
@@ -35,7 +35,7 @@ export const SkeletonText: FC<TextProps> = ({ children, ...rest }) => {
 
   return (
     <Flex alignSelf="flex-start">
-      <Text {...rest} bg={color("black10")} color={color("black10")}>
+      <Text {...rest} bg={color("black10")} color={color("rgba(0, 0, 0, 0)")}>
```
damassi, line 38: This breaks darkmode

src/elements/Screen/Header.tsx
```diff
@@ -82,14 +105,15 @@ const Center: React.FC<{
...
+  const { width: screenWidth } = useScreenDimensions()
 
   if (hideTitle) {
     return null
   }
 
   if (!animated) {
     return (
-      <Flex flex={1} flexDirection="row" width="100%">
+      <Flex position="absolute" left={0} flexDirection="row" minWidth={screenWidth}>
```
MounirDhahri, line 116: did position `absolute` with a defined `width` not work here?

damassi (changes requested):
I have a sneaking suspicion that this change might break Folio, which has a different manner of of title behavior than eigen (mostly left aligned, up to the width of the left element). Can we achieve the same thing by remaining with flexbox? As soon as things become manually positioned we lose all manner of flexibility. 

Also, not sure i'm seeing the issue in the screenshots.

MounirDhahri (approved):
LGTM, I am sure you already did it, but I would only suggest to test this in different screens because it's being used not only in Eigen screens but also in Folio

src/elements/Skeleton/Skeleton.tsx
olerichter00: Thanks! This change isn't releated to the fix, forgot to remove it before pushing my changes 🙄

olerichter00:
> I have a sneaking suspicion that this change might break Folio, which has a different manner of of title behavior than eigen (mostly left aligned, up to the width of the left element). Can we achieve the same thing by remaining with flexbox? As soon as things become manually positioned we lose all manner of flexibility.
Also, not sure i'm seeing the issue in the screenshots.

Thanks for the reviews 🙏 

Before this fix, we had to set the width of the right element manually to ensure it is rendered correctly ([Eigen code](https://github.com/artsy/eigen/blob/adada5761a116f314f21a0ce60453c616fd285a4/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx#L46)).

I just realized that what I'm trying to achieve with `position: absolute` would also work by setting `flex: 1` for both the left and right elements to ensure they have the same size, positioning the title exactly in the center of the screen ([JSFiddle with CSS example](https://jsfiddle.net/olerichter00/7ru59kzo/))💡 I'll update this PR and get rid of the `position: absolute`.

damassi:
Sweet, that seems like a good fix!

olerichter00:
Closing this PR in favor of https://github.com/artsy/palette-mobile/pull/203

PR #201, opened 2024-03-04
fix: make sure image comes on top of the skeleton

MounirDhahri:
## To try this out, install the latest beta and get the last canary from code-push, called `Blurhashes everywhere`. 
**Make sure to enable palette-everywhere and show blurhash feature flags**

### Description

This PR solves https://www.notion.so/artsy/IT-looks-as-if-the-first-page-only-have-blur-hash-for-artist-images-366bdf8b2751449a943bb55fd0805d17?pvs=4

**What we have now:** Currently,  we have an Image component that is being covered by a skeleton. Once the image is done loading, we set the skeleton opacity to 0, and the image component shows up **from under it**. 

**Iusse:** If the UI thread and JS thread are both busy, we can end up having the skeleton opacity getting to 0, while the image (already loaded but not visible yet), has not be drawn yet. In this case, only the image box, which is black shows, and gives the impression there was no blurhash loaded. 

**Suggested change:**
The image component, is no longer hidden behind the skeleton, it's actually on top of it, it is initially not visible, has opacity of 0, and shows up gradually when it's done loading

**Impact of the above change ⬆️:**  There is **always** a Blurhash rendered, it's just hidden when the image is there. This means that, we don't need to worry about how long the image will take to be drawn, because even if it takes a while, there is a blurhash below.

How it looks like now

**Before**

https://github.com/artsy/palette-mobile/assets/11945712/90ed2297-e384-4080-a16d-3b0f9e6ffc15

**After**

https://github.com/artsy/palette-mobile/assets/11945712/ec835237-2c45-4a12-b2ba-8b8694049d86

https://github.com/artsy/palette-mobile/assets/11945712/813f5dce-9373-40ec-b1ea-aea42f77c8e3

https://github.com/artsy/palette-mobile/assets/11945712/a3c3e5ce-e8b2-4955-b52d-172d84fa4c38

### DIFF
--- a/src/elements/Image/Image.tsx
+++ b/src/elements/Image/Image.tsx
@@ -1,5 +1,5 @@
-import { memo } from "react"
-import { PixelRatio } from "react-native"
+import { memo, useCallback } from "react"
+import { PixelRatio, View } from "react-native"
 import { Blurhash } from "react-native-blurhash"
 import FastImage, { FastImageProps } from "react-native-fast-image"
 import Animated, {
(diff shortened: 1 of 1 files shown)


olerichter00 (approved):
👏

MrSltun (approved):
⚡

artsyit:
:rocket: PR was released in [`v13.1.18`](https://github.com/artsy/palette-mobile/releases/tag/v13.1.18) :rocket:

PR #204, opened 2024-03-06
fix: progressbar crash in dev and broken in prod

gkartalis:
This PR resolves [PHIRE-683] 

### Description

Noticed that the progress bar is broken in prod in the onboarding, it was caused due to the rn bump in palette that we did a while ago ([diff here](https://github.com/artsy/palette-mobile/commit/fcf0b32fc8b5faa823da6819080a23d826bb438e#diff-f31c81aceb520986e64c9824ec29a0cbc561fea2fc707996cc1b4a0d4684acf6R30))

The TS error that I tried to fix on the diff above was because useAnimatedStyle expects a return type that is compatible with AnimatedStyleProp<ViewStyle | ImageStyle | TextStyle>. The width property in this case should be a number or a SharedValue<number> instead of a string.

Fixed this by converting the width value to a number before assigning it to the width property in the useAnimatedStyle hook.

This change assumes that the progress prop passed to the ProgressBar component is a number between 0 and 100. The width value is then calculated as a percentage of this progress as before.

### in eigen:

|Before|After|
|---|---|
|Screenshot 2024-03-06 at 13 52 56|Simulator Screenshot - iPhone 14 Pro - 2024-03-06 at 13 53 07|

### in palette: 
|Before|After|
|---|---|
|Simulator Screenshot - iPhone 15 - 2024-03-06 at 13 54 46|Simulator Screenshot - iPhone 15 - 2024-03-06 at 13 49 38|

[PHIRE-683]: https://artsyproduct.atlassian.net/browse/PHIRE-683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/elements/ProgressBar/ProgressBar.tsx
+++ b/src/elements/ProgressBar/ProgressBar.tsx
@@ -24,16 +24,16 @@ export const ProgressBar = ({
   trackColor = "blue100",
 }: ProgressBarProps) => {
   const color = useColor()
-  const width = useSharedValue("0%")
+  const width = useSharedValue(0)
   const progress = clamp(unclampedProgress, 0, 100)
   const progressAnim = useAnimatedStyle(() => {
-    return { width: Number(width.value) }
+    return { width: `${width.value}%` }
   })
 
   const [onCompletionCalled, setOnCompletionCalled] = useState(false)
 
   useEffect(() => {
-    width.value = withTiming(`${progress}%`, { duration: animationDuration })
+    width.value = withTiming(progress, { duration: animationDuration })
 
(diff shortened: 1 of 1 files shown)


MrSltun (approved):

artsyit:
:rocket: PR was released in [`v13.1.20`](https://github.com/artsy/palette-mobile/releases/tag/v13.1.20) :rocket:

MounirDhahri:
Thank you 🙏

PR #206, opened 2024-03-15
fix(radio): fix alignment between button and text

anandaroop:
This PR resolves [ONYX-819] 

### Description

Radio buttons in palette-mobile currently suffer from poor alignment between button and text. 

Example from SwA flow in Eigen:

It seems to me that people have worked around this problem in Eigen by building up custom flex layouts that include:

- a `<RadioButton />` _without_ the `text` prop
- combined with a custom `<Text />` element
- all of the above wrapped in a `<Touchable />`

(Examples [here](https://github.com/artsy/eigen/blob/7423ddbfa513f79a638193434a118bdc13261d95/src/app/Scenes/Artwork/Components/ArtworkEditionSetItem.tsx#L41-L64) and [here](https://github.com/artsy/eigen/blob/7423ddbfa513f79a638193434a118bdc13261d95/src/app/system/devTools/DevMenu/CodePushOptions.tsx#L81-L89) and [here](https://github.com/artsy/eigen/blob/7423ddbfa513f79a638193434a118bdc13261d95/src/app/Scenes/SavedSearchAlert/Components/SavedSearchFilterSize.tsx#L99-L114))

I think it would be better to fix this at the root, here in palette-mobile, so that these^ contortions are not necessary and the `RadioButton` can simply be used as-is, with the `text` prop.

That's what this PR does:

|Before|After|
|---|---|
|before|after|

There are only a handful of cases in Eigen and Energy where the component is used as-is. Those will just inherit the improved alignment once they upgrade to this version of palette-mobile.

The more custom layouts I cited above should not be affected — though it might be nice to simplify those components and use the improved behavior now.

[ONYX-819]: https://artsyproduct.atlassian.net/browse/ONYX-819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.storybook/storybook.requires.js
+++ b/.storybook/storybook.requires.js
@@ -46,6 +46,7 @@ const getStories = () => {
     require("../src/elements/Pill/Pill.stories.tsx"),
     require("../src/elements/Popover/Popover.stories.tsx"),
     require("../src/elements/ProgressBar/ProgressBar.stories.tsx"),
+    require("../src/elements/Radio/RadioButton.stories.tsx"),
     require("../src/elements/Screen/Screen.stories.tsx"),
     require("../src/elements/Separator/Separator.stories.tsx"),
     require("../src/elements/Skeleton/Skeleton.stories.tsx"),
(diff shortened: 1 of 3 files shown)


gkartalis (approved):
nice! thanks!

MounirDhahri (approved):
Thanks!

artsyit:
:rocket: PR was released in [`v13.1.22`](https://github.com/artsy/palette-mobile/releases/tag/v13.1.22) :rocket:

PR #208, opened 2024-03-25
feat: Add `preventUnmount` prop to `CollapsibleMenuItem`

olerichter00:
This PR resolves [ONYX-815] 

- Related Eigen PR: https://github.com/artsy/eigen/pull/9998

### Description

This PR adds a `preventUnmount` prop to `CollapsibleMenuItem` that prevents unmounting the content when collapsed. This change is needed in order to not lose the current state when a step in the SWA form is collapsed.

## Screenshots

Simulator Screenshot - iPhone 15 Pro - 2024-03-25 at 16 45 43

[ONYX-815]: https://artsyproduct.atlassian.net/browse/ONYX-815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/elements/Collapse/Collapse.tsx
+++ b/src/elements/Collapse/Collapse.tsx
@@ -5,7 +5,7 @@ interface CollapseProps {
   opened: boolean
   children: ReactNode
   /**
-   * Prevents the component from unmounting when it is closed  (display: none)
+   * Prevent the content from being unmounted when it is collapsed (display: none)
    */
   preventUnmount?: boolean
 }
--- a/src/elements/Collapse/Collapse.tsx
+++ b/src/elements/Collapse/Collapse.tsx
@@ -1,5 +1,17 @@
 import { ReactNode } from "react"
 import { View } from "react-native"
 
-export const Collapse = ({ opened, children }: { opened: boolean; children: ReactNode }) =>
-  opened ? <View>{children}</View> : null
+interface CollapseProps {
+  opened: boolean
+  children: ReactNode
+  /**
+   * Prevents the component from unmounting when it is closed  (display: none)
+   */
+  preventUnmount?: boolean
+}
+
+export const Collapse: React.FC<CollapseProps> = ({ opened, children, preventUnmount = false }) => {
+  if (preventUnmount) return <View style={{ display: opened ? null : "none" }}>{children}</View>
+
+  return opened ? <View>{children}</View> : null
+}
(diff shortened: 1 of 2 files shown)


src/elements/Collapse/Collapse.tsx
```diff
@@ -1,5 +1,18 @@
...
+  opened: boolean
+  children: ReactNode
+  /**
+   * Prevent the content from being unmounted when it is collapsed (display: none)
+   */
+  preventUnmount?: boolean
+}
+
+export const Collapse: React.FC<CollapseProps> = ({ opened, children, preventUnmount = false }) => {
+  if (preventUnmount)
```
nickskalkin, line 14: **question / non-blocking**: I am wondering if we ever need "unmountable" behaviour. Maybe toggling `display` (none/undefined) property based on `opened` argument is enough?

nickskalkin (approved):

src/elements/Collapse/Collapse.tsx
olerichter00: Good question 🤔  I think it should be okay, but I was wondering if there could be (performance) issues and was too lazy to think about testing things. @MounirDhahri, what are your thoughts on this?

If there are no implications to expect, then yeah, let's go with your suggestion.

MounirDhahri: I have the same thought as nikita here, i think it's safe to keep the content as well. I would though suggest to test things out in the contact gallery modal.

src/elements/Collapse/Collapse.tsx
```diff
@@ -1,5 +1,18 @@
...
+    return <View style={{ display: opened ? undefined : "none" }}>{children}</View>
```
MounirDhahri, line 15: I understand this means keeping things as they are, I would suggest though seeing if we can animate how this content shows up

src/elements/Collapse/Collapse.tsx
olerichter00: Alright, let's remove the prop and always keep the content unmounted 👍 

I will test things in the app and also write a note for the Recent Changes QA session.

artsyit:
:rocket: PR was released in [`v13.1.23`](https://github.com/artsy/palette-mobile/releases/tag/v13.1.23) :rocket:

PR #210, opened 2024-04-02
fix: add zIndex to the ToolTip component

dariakoko:
This PR resolves [] 

### Description

This PR fixes the ToolTip component.
In some implementations the content under the ToolTip is visible - we should avoid that.
Fixed by adding zIndex to the ToolTipFlyout 

| Before | After |
| --- | --- |
| image | image |


artsyit:
:rocket: PR was released in [`v13.1.24`](https://github.com/artsy/palette-mobile/releases/tag/v13.1.24) :rocket:

PR #215, opened 2024-04-17
fix(APPL-872): return null if Collapse is not opened

MrSltun:
This PR resolves [APPL-872] 

### Description
This PR updates the Collapse element to not render a view when the Collapse is not opened

cc @artsy/onyx-devs 

[APPL-872]: https://artsyproduct.atlassian.net/browse/APPL-872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/elements/Collapse/Collapse.tsx
+++ b/src/elements/Collapse/Collapse.tsx
@@ -7,5 +7,9 @@ export interface CollapseProps {
 }
 
 export const Collapse: React.FC<CollapseProps> = ({ opened, children }) => {
-  return <View style={{ display: opened ? undefined : "none" }}>{children}</View>
+  if (!opened) {
+    return null
+  }
+
+  return <View>{children}</View>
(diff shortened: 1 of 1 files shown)


brainbicycle (approved):
👍 thank you!

gkartalis (approved):

brainbicycle:
just a heads up to onyx this seems to work fine but does make the transition a bit jumpy. It fixes the worse issue of hidden elements displaying and causing the app to become unresponsive though.

olerichter00:
Thanks for the fix! I will look into this again, because we've made a change to this component to not unmount the content when it's collapsed. This change was supposed to keep the state when going back to a previous step in the SWA flow.

PR #225, opened 2024-04-30
feat: support placeholders in Input

MounirDhahri:
### Description

This PR comes as a follow-up on https://github.com/artsy/eigen/pull/10167
Here we update the Input component to support arrays - this is in order to fix an issue on android where the input placeholder is not being calculated properly.

[sorry the uploaded didn't work on github]
https://artsy.slack.com/archives/C02BAQ5K7/p1714492481009039?thread_ts=1714138172.001059&cid=C02BAQ5K7

### DIFF
--- a/src/elements/Input/Input.tsx
+++ b/src/elements/Input/Input.tsx
@@ -2,8 +2,8 @@ import { EventEmitter } from "events"
 
 import { THEME } from "@artsy/palette-tokens"
 import themeGet from "@styled-system/theme-get"
-import isString from "lodash/isString"
 import isArray from "lodash/isArray"
+import isString from "lodash/isString"
 import {
   RefObject,
   forwardRef,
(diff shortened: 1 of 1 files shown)


src/elements/Input/Input.tsx
```diff
@@ -551,8 +602,42 @@ export const Input = forwardRef<InputRef, InputProps>(
...
+            alignItems: "baseline", // this is to make Texts get the smallest width they can get to fit the text
+          }}
+        >
+          {placeholder.map((placeholderString, index) => (
+            <Text
+              onLayout={(event) => {
+                placeholderWidths.current[index] = event.nativeEvent.layout.width
+              }}
+              numberOfLines={1}
+              style={{
```
araujobarret, line 625: nit: style={styles}

src/elements/Input/Input.tsx
```diff
@@ -2,6 +2,7 @@ import { EventEmitter } from "events"
 
 import { THEME } from "@artsy/palette-tokens"
 import themeGet from "@styled-system/theme-get"
+import _ from "lodash"
```
araujobarret, line 5:
```suggestion
import isString from "lodash/isString"
import isArray from "lodash/isArray"
```

src/elements/Input/Input.tsx
```diff
@@ -551,8 +602,42 @@ export const Input = forwardRef<InputRef, InputProps>(
       )
     }, [labelStyles, labelAnimatedStyles, props.title])
 
+    const renderAndroidPlaceholderMeasuringHack = useCallback(() => {
+      if (Platform.OS === "ios" || !_.isArray(placeholder)) {
```
araujobarret, line 606: I vote for when the placeholder is measured we don't render it anymore, just add the extra check here to return null in that case

src/elements/Input/Input.tsx
```diff
@@ -551,8 +602,42 @@ export const Input = forwardRef<InputRef, InputProps>(
...
+              style={{
+                ...styles,
+              }}
+            >
+              {placeholderString}
+            </Text>
+          ))}
+        </Flex>
+      )
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
araujobarret, line 634: let's use all the deps the hook is asking for 🙇

src/elements/Input/Input.tsx
MounirDhahri: Because RN bundles everything anyway I figured out this is not needed here - I can change that though as a good practice to keep from Web.

src/elements/Input/Input.tsx
MounirDhahri: good idea - will do that ;)

araujobarret (approved):

artsyit:
:rocket: PR was released in [`v13.2.3`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.3) :rocket:

PR #230, opened 2024-05-27
chore: support progress bar styles

MounirDhahri:
Support progress bar styles

This is required because we always inject a margin that is not always necessary.


artsyit:
:rocket: PR was released in [`v13.2.8`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.8) :rocket:

PR #232, opened 2024-06-05
chore: allow disabling keyboard avoidance

MounirDhahri:
This PR resolves [] 

### Description

This PR adds support to disabling keyboard avoidance to `Screen.Body`

**Why am I adding this?**
When the user dismisses a screen where they had the keyboard visible, the `KeyboardAvoidingView` takes action of adjusting the screen content in order to fit it within the screen. Most of the cases, this will result in nothing, since there is nothing to push up, however in the case where we have a sticky button (like the case of SWA landing page "Start Selling" button), we will see that button going down along with the keyboard. Which is something you wouldn't expect since the screen wasn't active.

https://github.com/artsy/palette-mobile/assets/11945712/15372521-dd88-4bc5-8ea0-dd0fb9bec8fb

### DIFF
--- a/src/elements/Screen/Body.tsx
+++ b/src/elements/Screen/Body.tsx
@@ -1,3 +1,4 @@
+import { Fragment } from "react"
 import { KeyboardAvoidingView, Platform, ScrollView } from "react-native"
 import { SCREEN_HORIZONTAL_PADDING } from "./constants"
 import { Flex, FlexProps } from "../Flex"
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in [`v13.2.10`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.10) :rocket:

dariakoko:
Thank you Mounir!

PR #233, opened 2024-06-05
chore(Body): do not inject styles in Fragment component

MounirDhahri:
### Description

This PR comes to address a potential prop leak introduced in the previous PR https://github.com/artsy/palette-mobile/pull/232

### DIFF
--- a/src/elements/Screen/Body.tsx
+++ b/src/elements/Screen/Body.tsx
@@ -1,5 +1,4 @@
-import { Fragment } from "react"
-import { KeyboardAvoidingView, Platform, ScrollView } from "react-native"
+import { KeyboardAvoidingView, ScrollView } from "react-native"
 import { SCREEN_HORIZONTAL_PADDING } from "./constants"
 import { Flex, FlexProps } from "../Flex"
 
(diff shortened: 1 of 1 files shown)


src/elements/Screen/Body.tsx
```diff
@@ -16,15 +15,17 @@ export const Body: React.FC<BodyProps> = ({
...
+    <Flex flex={1} {...flexProps}>
+      <Flex flex={1} px={fullwidth ? undefined : SCREEN_HORIZONTAL_PADDING}>
+        {scroll ? <ScrollView>{children}</ScrollView> : children}
       </Flex>
-    </Wrapper>
+    </Flex>
   )
+
+  if (disableKeyboardAvoidance) {
+    return Content
```
MounirDhahri, line 27: I hate conditional wrappers 😄 there is not beautiful way to write them that I know of

gkartalis (approved):
thanks

src/elements/Screen/Body.tsx
gkartalis: hmm actually what do you think about something like this:

gkartalis:
```
<Wrapper
        {...(disableKeyboardAvoidance
          ? {}
          : { style: { flex: 1 }, behavior: Platform.OS === "ios" ? "padding" : "height" })}
      >
```

With this way we are getting away without conditional wrappers

MounirDhahri: not the sexiest but would also work - will update

MounirDhahri:
thanks to you 🌷

artsyit:
:rocket: PR was released in [`v13.2.11`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.11) :rocket:

PR #235, opened 2024-06-06
chore: support progressBarStyle for ProgressBar component

MounirDhahri:
### Description

This PR adds support of injecting styles into the progress bar component. Although ideally we should not inject any styles here, Design wants to be able adjust a few properties of the progress bar when it's not full screen. In the most recent case, making it round

simulator_screenshot_9A4082D6-62AD-4E75-9BE7-327DD2E2ED49


artsyit:
:rocket: PR was released in [`v13.2.13`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.13) :rocket:

PR #243, opened 2024-07-03
feat(Button): respect variant style given loading state

araujobarret:
### Description

I realised we never moved out to use the ButtonNew(which is neat and has the loading state respecting the variant).
So, this PR brings the stories to the Button we use and uses the property styling given loading state with all variants supported by it.

Before:

https://github.com/artsy/palette-mobile/assets/15792853/e738fbaf-c437-4229-9c08-db96aa29c8ed

After:

https://github.com/artsy/palette-mobile/assets/15792853/c826e479-8109-4e82-acef-5a79a0ff3a07


artsyit:
:rocket: PR was released in [`v13.2.21`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.21) :rocket:

PR #246, opened 2024-07-25
fix: multiline input scrolling

iskounen:
This PR addresses a bug where scrolling was always disabled for input fields, including multiline ones. The result is that none of the multiline inputs in Eigen can display more than 3 lines of text to users.

| Before | After |
|-------|-------|
| scroll-disabled | scroll-enabled |


MounirDhahri (approved):
Nice catch! thanks for fixing this Adam

MrSltun (approved):
Thank you Adam! 🌟 🙌

artsyit:
:rocket: PR was released in [`v13.2.24`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.24) :rocket:

PR #249, opened 2024-09-09
feat(Button): swaps the current button by the new button using reanimated

araujobarret:
### Description

- Removes the old button
- moves the new button elements to the button folder structure
- removes the element LinkButton which was basically a Text element with Touchable and callback, we don't need this abstraction since in Eigen we use links in black60, black100, and we use LinkButton only in one file in Eigen 

Follow-ups:
- I noticed the text from the FollowButton doesn't change its color when pressed, we tried some tweaks but needs to be refactored in order to fully work
- icons need to be refactored if we want to change the color when pressed

I have the code in Eigen already in place, once this PR is merged I will open the PR there to bump palette.


src/elements/Button/Button.stories.tsx
```diff
@@ -80,21 +88,146 @@ export const Variants = () => (
...
+          </Wrap.Content>
         </Flex>
-      )
-    }}
+      </Wrap>
+    )}
   />
 )
+
+export const TheFollowButton = () => {
```
araujobarret, line 116: the diff of some files is because I moved from ButtonNew to Button basically

src/elements/Button/LinkButton.tsx
```diff
@@ -1,8 +0,0 @@
-import { TextProps, Text } from "../Text"
-import { Touchable } from "../Touchable"
-
-export const LinkButton = (props: TextProps) => (
-  <Touchable onPress={props.onPress}>
-    <Text underline {...props} />
-  </Touchable>
-)
```
gkartalis, line 8: The `<LinkButton />` is being consumed by eigen [here](https://github.com/artsy/eigen/blob/61fb6e3a57cbcdcae22187d490626fa4e9427cb5/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddPhotos.tsx#L46-L54) could you address that since it is a breaking change?

araujobarret: it makes sense, I think I'll go ahead and open the Eigen PR and then merge this one, I don't think this one needs to be merged first 👍

gkartalis:
@araujobarret since you did it locally then it should be good  👍

MrSltun (approved):
LGTM! 🌟 
Thank you Carlos! 🙌

gkartalis (approved):

artsyit:
:rocket: PR was released in [`v13.2.27`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.27) :rocket:

PR #253, opened 2024-09-16
refactor(Button): enable loading state for button

araujobarret:
### Description

Fixes loading states similar to what was done in #243. 
Co-authored by @gkartalis 

https://github.com/user-attachments/assets/40794982-fd80-449f-8a96-517e172cf934

### DIFF
--- a/src/elements/Button/Button.tsx
+++ b/src/elements/Button/Button.tsx
@@ -117,8 +117,6 @@ export const Button = ({
     }
   })()
 
-  const spinnerColor = variant === "text" ? "blue100" : "white100"
-
   const handlePress = (event: GestureResponderEvent) => {
     if (onPress === undefined || onPress === null) {
       return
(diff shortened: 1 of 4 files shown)


src/elements/Button/Button.tsx
```diff
@@ -213,46 +221,42 @@ export const Button = ({
...
+              {iconPosition === "right" && !!icon && (
                 <>
                   <Spacer x={0.5} />
                   {icon}
                 </>
-              ) : null}
+              )}
 
-              {loading ? (
+              {loading && (
```
gkartalis, line 249: 🤔 can't believe that this was the issue 🤔

gkartalis (approved):

src/elements/Button/Button.tsx
araujobarret: it wasn't we we missing the loader colors and the transparencies to the text too

gkartalis: I get that the colors were missing but what about the size of the button? It is not affected by the colors right?

araujobarret:
Added a missing fix on the follow button here https://github.com/artsy/palette-mobile/pull/253/commits/1929b176415957b76225f13fcc4c32fa78316f59

result:

https://github.com/user-attachments/assets/21c308b9-6377-459a-b443-4c5182aabf12

gkartalis (approved):

artsyit:
:rocket: PR was released in [`v13.2.30`](https://github.com/artsy/palette-mobile/releases/tag/v13.2.30) :rocket:

PR #275, opened 2024-10-31
feat: add new heart and follow icons

dariakoko:
This PR resolves [ONYX-1286] [ONYX-1293] 

### Description

To address https://artsyproduct.atlassian.net/browse/ONYX-1286 and https://artsyproduct.atlassian.net/browse/ONYX-1293 we need to add new icons
image
image

[ONYX-1286]: https://artsyproduct.atlassian.net/browse/ONYX-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ONYX-1293]: https://artsyproduct.atlassian.net/browse/ONYX-1293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/svgs/NewFillHeartIcon.tsx
```diff
@@ -0,0 +1,16 @@
+import { Icon, IconProps, Path } from "./Icon"
+import { useColor } from "../utils/hooks"
+
+export const NewFillHeartIcon = ({ fill, ...restProps }: IconProps) => {
```
dariakoko, line 4: NewFillHeartIcon and NewHeartIcon are not the perfect names...
After [the feature](https://artsyproduct.atlassian.net/browse/ONYX-1286) is release and corresponding feature flag is no longer in use we can adjust the FillHeartIcon and HeartIcon components and remove those added in this PR

src/svgs/FollowArtistFillIcon.tsx
```diff
@@ -0,0 +1,16 @@
+import { Icon, IconProps, Path } from "./Icon"
+import { useColor } from "../utils/hooks"
+
+export const FollowArtistFillIcon = ({ fill, ...restProps }: IconProps) => {
+  const color = useColor()
+
+  return (
+    <Icon {...restProps} viewBox="0 0 17 13">
```
gkartalis, line 8: 🤔 agree with @araujobarret point around the screenshots, if I remember correctly almost all of our icons have `viewBox="0 0 18 18"` screenshots would be awesome in order to see if these look right or not.

I think you don't need to add a story, I think it is already handled and maps over all of the icons

dariakoko: Addressed both comments, thanks!

gkartalis (approved):
thank you!

PR #279, opened 2024-11-12
feat: animate icons

MounirDhahri:
### Description
When I was on WhatsApp, I noticed this nice transition - so I also rebuilt it in our search input component

https://github.com/user-attachments/assets/8518a518-d9cd-45e8-bd7d-2517a0454e26

https://github.com/user-attachments/assets/2cb25500-661e-4e27-89c4-b05f5ad28fd4


dariakoko:
Let's raise the topic of icons animation at product-onyx? We talked about animating the Heart icons before as well, could be an interesting improvement

PR #280, opened 2024-11-14
[Do Not Merge] => feat(expo): Migrate app to Expo

damassi:
### Description

Similar to the [Folio Expo migration](https://github.com/artsy/energy/pull/1196) this spikes on migrating palette mobile over as well. 

TLDR: The toolchain is legit! I'm honestly kinda amazed by all of the things that are packed into `npx expo start`, and impressed at how easy it was to get all of the deps fully configured with `npx expo-doctor`. 

Given the abundant docs, I took a couple right turns initially, but then started over and everything worked very painlessly and was able to set up from scratch and run it in the iOS simulator in about 15 minutes. 

Here are the steps I took:

- Install and auto-configure Expo: `npx install-expo-modules@latest` 
- Run `npx expo-doctor`. Noticed there were a few package versions that weren't aligned with expo, a few minor versions away
- Run `npx expo install --check`. Installed everything, automatically updated package.json
- Update metro config with expo-centric settings
```tsx
const { getDefaultConfig } = require("@expo/metro-config")
const config = getDefaultConfig(__dirname)
module.exports = config
```
- Remove `main` entry in package.json. This really stumped me initially before I bailed on trying to work around it. Expo reads this as the entrypoint for the app and it **is not negotiable**! There is no more `entryPoint` on app.json. Previously this pointed at the `dist/index.js`, the npm-publish friendly (compiled) version of our consumable `@artsy/palette-mobile` lib. Need to test whether lib still works in consumers without main. I think its fine. 
- Compile and boot iOS: `npx expo run:ios` 
- Compile and android: `npx expo run:android` 
- Start in dev mode: `npx expo start` 
- App boots and runs great

Current status:

I'm currently testing completely deleting preexisting `android` and `ios` folders to see how expo's fully managed config works. iOS seemed to run into an issue, but i suspect its trivial. Android is still compiling.


package.json
```diff
@@ -118,19 +117,19 @@
...
     "react-dom": "18.2.0",
-    "react-native": "0.75.4",
+    "react-native": "0.74.5",
     "react-native-device-info": "10.3.0",
     "react-native-haptic-feedback": "1.14.0",
     "react-native-linear-gradient": "2.6.2",
-    "react-native-reanimated": "3.16.1",
-    "react-native-safe-area-context": "4.11.1",
-    "react-native-svg": "14.1.0",
+    "react-native-reanimated": "~3.10.1",
```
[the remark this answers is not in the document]
damassi: I completely agree!

ios/PaletteMobile/PrivacyInfo.xcprivacy
```diff
@@ -4,20 +4,31 @@
...
 		</dict>
 		<dict>
 			<key>NSPrivacyAccessedAPIType</key>
-			<string>NSPrivacyAccessedAPICategoryUserDefaults</string>
+			<string>NSPrivacyAccessedAPICategoryDiskSpace</string>
 			<key>NSPrivacyAccessedAPITypeReasons</key>
 			<array>
-				<string>CA92.1</string>
+				<string>E174.1</string>
+				<string>85F4.1</string>
```
brainbicycle, line 31: this is maybe just because palette-mobiles privacy manifest wasn't accurate? but it might also be a side effect of pulling in expo which is a bit unfortunate, these show up on our privacy report card in App Store so adding extra keys is not ideal. Not the end of the world but not great.

ios/PaletteMobile.xcodeproj/project.pbxproj
```diff
@@ -472,7 +515,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 12.4;
+				IPHONEOS_DEPLOYMENT_TARGET = 13.4;
```
brainbicycle, line 518: i wonder if we have control over this? we are currently higher in eigen. not a deal breaker but something to look out for as these little keys have big consequences.

ios/PaletteMobile/PrivacyInfo.xcprivacy
damassi: I bet we can tweak this -- these were added when expo ran the first pass auto-fix via 
```
npx install-expo-modules@latest
```

I didn't take this PR all the way to completion (notably things like assets and the like configure via app.json), so reminder this was just a first pass to determine ease and to get a feel for the process. 

If we want to proceed I'd want to take it one step further:
- free plan EAS builds 
- delete ios and android directories 
- Finish all of the static configuration so that we don't have to worry about asset configuration (splash, etc)

android/app/src/main/res/values/styles.xml
```diff
@@ -1,9 +1,17 @@
...
+    <item name="android:statusBarColor">#ffffff</item>
+  </style>
+  <style name="ResetEditText" parent="@android:style/Widget.EditText">
+    <item name="android:padding">0dp</item>
+    <item name="android:textColorHint">#c8c8c8</item>
+    <item name="android:textColor">@android:color/black</item>
+  </style>
+  <style name="Theme.App.SplashScreen" parent="AppTheme">
+    <item name="android:windowBackground">@drawable/splashscreen</item>
+  </style>
```
brainbicycle, line 16: does expo replace react-native-bootsplash in eigen's case?

android/app/src/main/java/com/palettemobile/MainApplication.kt
```diff
@@ -1,43 +0,0 @@
...
-              // Packages that cannot be autolinked yet can be added manually here, for example:
-              // add(MyReactNativePackage())
-            }
-
-        override fun getJSMainModuleName(): String = "index"
-
-        override fun getUseDeveloperSupport(): Boolean = BuildConfig.DEBUG
-
-        override val isNewArchEnabled: Boolean = BuildConfig.IS_NEW_ARCHITECTURE_ENABLED
-        override val isHermesEnabled: Boolean = BuildConfig.IS_HERMES_ENABLED
```
brainbicycle, line 29: not important for palette but a call out if we do move to eigen we have customization in these native files for configuring some deps, I think Braze is one. We should try to understand what we have to do to either keep a custom version or generate a custom version or probably a lot of libraries have migration instructions.

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,25 +1,33 @@
...
-            <category android:name="android.intent.category.LAUNCHER" />
-        </intent-filter>
-      </activity>
-    </application>
-</manifest>
+  <uses-permission android:name="android.permission.INTERNET"/>
+  <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
+  <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
+  <uses-permission android:name="android.permission.VIBRATE"/>
+  <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
```
brainbicycle, line 6: seems to have added a bunch of permissions, not a concern for palette-mobile, probably not a concern for eigen unless they differ from what is already there? but we should look out for it

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,25 +1,33 @@
...
+  <queries>
+    <intent>
+      <action android:name="android.intent.action.VIEW"/>
+      <category android:name="android.intent.category.BROWSABLE"/>
+      <data android:scheme="https"/>
```
brainbicycle, line 11: have some custom domains in eigen for deeplinking, would be good to understand how that works in expo

android/app/proguard-rules.pro
```diff
@@ -7,4 +7,8 @@
 # For more details, see
 #   http://developer.android.com/guide/developing/tools/proguard.html
 
+# react-native-reanimated
+-keep class com.swmansion.reanimated.** { *; }
+-keep class com.facebook.react.turbomodule.** { *; }
```
brainbicycle, line 12: oh nice maybe this makes proguard adoption easier?

ios/PaletteMobile/PrivacyInfo.xcprivacy
brainbicycle: eigen has more in here so really only matters if there is a diff between what is there now and what we get by bringing in expo but something to look out for

package.json
```diff
@@ -118,19 +117,19 @@
     "postinstall-prepare": "^2.0.0",
     "prettier": "^2.8.8",
     "pull-lock": "1.0.0",
-    "react": "18.3.1",
+    "react": "18.2.0",
     "react-dom": "18.2.0",
-    "react-native": "0.75.4",
+    "react-native": "0.74.5",
```
brainbicycle, line 122: this is one of my bigger concerns, locking us to specific dependencies. we fairly often have to upgrade things to workround issues and from what I can see here:


downgraded packages

-  react native downgraded to 0.74.5 from 0.75.4
- react native reanimated to 3.10.1 from 3.16.1
- react from 18.2.0 to 18.3.1
- react-native-safe-area-context from 4.11.1 to 4.10.5
- @shopify/flash-list from 1.7.1 to 1.6.4
- react-native-pager-view from 6.5.0 to 6.3.0

upgraded packages:
- react-native-svg from 14.1.0 to 15.2.0
- typescript from 5.0.4 to 5.3.3
- @react-native-async-storage/async-storage 1.19.8 to 1.23.1

brainbicycle: if we have a major bug that we need to upgrade for it would be good to understand what our options are, does that mean we have to eject? or patching is our only option?

ios/PaletteMobile/PrivacyInfo.xcprivacy
brainbicycle: yeah I know, these comments are me trying to get more of a read on what this would take in Eigen than a typical PR review

damassi: Oh totally understand. Just clarifying that what's seen here is technically incomplete 👍

package.json
brainbicycle: maybe I will check also how often this is actually true `we fairly often have to upgrade things to workround issues` I feel like it is true but should be able to see in pr history

gkartalis:
@MounirDhahri @damassi a thing that I noticed in the community for react native libraries that use expo:

Seems like people are leaving expo outside the main library but they use it on an example app for running the library as an example there and the experience is pretty awesome and really fast + easy to contribute:

- https://github.com/gorhom/react-native-bottom-sheet
- https://github.com/PedroBern/react-native-collapsible-tab-view

seems like we might need to treat palette-mobile a bit differently when it comes to an expo migration and treat it as a dependency and maybe have an example app that includes expo to be able to run it and develop on it taking the advantages of expo while at the same time not bloating the actual library code with extra deps.

Happy to hear your thoughts on this.

package.json
MounirDhahri: fwiw, we can still keep the packages to the versions we want to and we don't have to be locked to what `expo --fix` suggests. The upgrades/downgrades are just suggestions we can ignore if we want to.

gkartalis:
@damassi note that also the libraries that I did mention do not include any android + ios folders since they are js only react native libraries but they do not include expo as a dep on their package.json.

We could have something similar - like keeping the components on the `src` folder or something and have an example app - which can be an expo app and from there we can run an expo storybook app (or even a normal app without storybook with component examples)

package.json
```diff
@@ -34,20 +34,24 @@
...
     "react-native-collapsible-tab-view": "^8.0.0",
     "react-native-fast-image": "^8.6.3",
-    "react-native-pager-view": "6.5.0",
+    "react-native-pager-view": "6.3.0",
     "react-native-popover-view": "^5.1.7",
     "react-spring": "8.0.23",
-    "styled-system": "^5.1.5"
+    "styled-system": "^5.1.5",
+    "react": "18.2.0",
+    "react-native": "0.74.5"
```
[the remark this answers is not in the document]
brainbicycle: yeah it sounds like it should!

PR #281, opened 2024-11-18
chore: allow using uncontrolled inputs

MounirDhahri:
### Description

When trying to log in using the new flow, I noticed some keystrokes not making it. I did some investigation to see why this is happening since I was assuming we are using a non-controlled input for it so performance should not ideal. However, I noticed that:

1. We accidentally brought back the `value` prop [here](https://github.com/artsy/eigen/blob/323d364c69ac26ba67050672cbefaddddad5a075/src/app/Scenes/Onboarding/Auth2/scenes/ForgotPasswordStep.tsx#L123)
2. The value is being set in Palette Mobile, regardless of using it or not 😮 

Here I am trying my artsy mail mounirdhahri[at]artsymail.com
1. In the video on the left, you can see how it's missing some keystrokes
2. In the video on the right, it's working smoothly (apart from the few mistakes I made 😄 )

| When using a controlled input | When using a non-controlled input |
|--------|--------|
| <Video src="https://github.com/user-attachments/assets/b814bfaf-7adb-4b1b-8620-a2fc6df6480c" /> | <Video src="https://github.com/user-attachments/assets/f241e7ab-5ac2-4f6d-8509-ed1963c9b799" />
 |

### DIFF
--- a/src/elements/Input/Input.tsx
+++ b/src/elements/Input/Input.tsx
@@ -49,6 +49,10 @@ export interface InputProps extends Omit<TextInputProps, "placeholder" | "onChan
    * These lead to some issues when the parent component wants further control of the value
    */
   disabled?: boolean
+  /**
+   * Enables the clear button
+   * @warning This prop only works if `value` is specified
+   */
   enableClearButton?: boolean
   error?: string
   fixedRightPlaceholder?: string
--- a/src/elements/Input/Input.tsx
+++ b/src/elements/Input/Input.tsx
@@ -98,6 +98,13 @@ export interface InputProps extends Omit<TextInputProps, "placeholder" | "onChan
    * <Input mask="999-99-9999 999-99-9999 999-99-9999" />
    */
   mask?: string | string[] | undefined
+  /**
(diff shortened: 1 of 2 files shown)


src/elements/Input/Input.tsx
```diff
@@ -98,6 +102,13 @@ export interface InputProps extends Omit<TextInputProps, "placeholder" | "onChan
    * <Input mask="999-99-9999 999-99-9999 999-99-9999" />
    */
   mask?: string | string[] | undefined
+  /**
+   * @warning This prop affects the performance of the input
+   * and should be avoided if possible.
+   * Use `defaultValue` instead.
+   * See: https://github.com/facebook/react-native-website/pull/4247
```
MounirDhahri, line 109: I highly encourage you to see the video in the issue to understand why I am adding this warning

damassi (approved):
🙏

artsyit:
:rocket: PR was released in [`v14.0.4`](https://github.com/artsy/palette-mobile/releases/tag/v14.0.4) :rocket:

PR #284, opened 2024-11-19
refactor: proper use of useSharedValue

gkartalis:
This PR resolves [PHIRE-1347] 

### Description

After upgrading reanimated we noticed a bunch of new warnings coming up.

They pointed to a missuse of `useSharedValue`, as seen [here](https://docs.swmansion.com/react-native-reanimated/docs/core/useSharedValue/#react-compiler-support) we need to avoid accessing the `.value` directly and also mutating values directly `.value = "whatever"`.

Instead we should use the getters and setters as pointed in the PR below.

Nothing changes in the functionality of the animations.

Some other bonus stuff:

properly enabled the `eslint-plugin-testing-library` that wasn't working properly before, and fixed some test issues.

[PHIRE-1347]: https://artsyproduct.atlassian.net/browse/PHIRE-1347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,7 @@ module.exports = {
     "plugin:import/typescript",
     "plugin:react-hooks/recommended",
     "plugin:react/jsx-runtime",
+    "plugin:testing-library/react",
     "prettier", // "prettier" needs to be last!
   ],
   parserOptions: {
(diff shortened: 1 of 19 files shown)


MounirDhahri (approved):
Thanks for this update. I must admit I am not a fan with the new API but I am sure they have their reasons

MrSltun (approved):
Thank you! 🌟

artsyit:
:rocket: PR was released in [`v14.0.4`](https://github.com/artsy/palette-mobile/releases/tag/v14.0.4) :rocket:

PR #293, opened 2024-12-06
chore: error when requesting resize for a gemini url 

brainbicycle:
This PR resolves [PHIRE-1391] 

### Description

Adds a loud warning when a gemini url is passed to the function to create gemini urls: https://artsy.slack.com/archives/C083MQTTR38/p1733414586530289?thread_ts=1733413352.174349&cid=C083MQTTR38

image-fail

[PHIRE-1391]: https://artsyproduct.atlassian.net/browse/PHIRE-1391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brainbicycle:
gonna go ahead and merge since a pretty trivial once, and would be nice to catch these early especially if we turn off the backend support again, but let me know if any concerns!

artsyit:
:rocket: PR was released in [`v14.0.12`](https://github.com/artsy/palette-mobile/releases/tag/v14.0.12) :rocket:

MounirDhahri:
@brainbicycle Can we actually have `performResize:false` as the default prop? I can't think of many cases where one would want to perform resize except for the images carousel

brainbicycle:
> @brainbicycle Can we actually have `performResize:false` as the default prop? I can't think of many cases where one would want to perform resize except for the images carousel

I think it makes sense but was opting into the lowest touch change possible because changing that is essentially changing the behavior of all images in our app and we have quite a few big changes happening at once already.

MounirDhahri:
Clear - thanks.

src/utils/createGeminiUrl.ts
```diff
@@ -17,6 +22,13 @@ export function createGeminiUrl({
   imageQuality?: number
   resizeMode?: GeminiResizeMode
 }) {
+  if (geminiHosts.some((host) => imageURL.includes(host))) {
```
MounirDhahri, line 25: I got this today.
Simulator Screenshot - iPhone 16 Pro - 2024-12-18 at 08 56 20

MounirDhahri: maybe we can be more defensive here

PR #294, opened 2024-12-13
chore: use isTestEnvironment instead of __test__

MounirDhahri:
This PR resolves [] 

### Description

This PR updates Palette to use `isTestingEnvironment()` instead of `__TEST__`. This is required for the Expo migration. 

This will be working for both Eigen and Energy so it shouldn't break anything

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,3 @@
-global.__TEST__ = false
-
 import { AppRegistry } from "react-native"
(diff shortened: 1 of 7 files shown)


artsyit:
:rocket: PR was released in [`v14.0.22`](https://github.com/artsy/palette-mobile/releases/tag/v14.0.22) :rocket:

PR #307, opened 2025-02-13
chore: fix dark mode tokens

MounirDhahri:
### Description
More Darkmode fixes on Folio

https://github.com/user-attachments/assets/b62b64ca-d9e0-4f3c-9b09-b766f2ed3d1e

### DIFF
--- a/src/tokens.ts
+++ b/src/tokens.ts
@@ -6,6 +6,7 @@
 
 import { THEME } from "@artsy/palette-tokens"
 import { SpacingUnit as SpacingUnitNumbers } from "@artsy/palette-tokens/dist/themes/v3"
+import { THEME_DARK } from "@artsy/palette-tokens/dist/themes/v3Dark"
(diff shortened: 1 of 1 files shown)


gkartalis (approved):

src/tokens.ts
```diff
@@ -109,7 +110,8 @@ const v3light: ThemeWithDarkModeType = {
 const v3dark: ThemeWithDarkModeType = {
   ...v3,
   colors: {
-    ...v3.colors,
+    ...THEME_DARK.colors,
```
MounirDhahri, line 113: we were using the wrong values here. We didn't notice it because of the time, we are using `background` `surface` for colours in Folio

artsyit:
:rocket: PR was released in [`v14.0.25`](https://github.com/artsy/palette-mobile/releases/tag/v14.0.25) :rocket:

PR #318, opened 2025-03-21
fix(Input): Input reset values on Android

araujobarret:
### Description

#305 introduced some regressions when you type on Android the input text gets flushed out. This PR adds a bandaid for that. 

I'd ask the reviewers to test on iPhone and Android if possible please 🙇 

Before:
 
https://github.com/user-attachments/assets/69327ec9-5804-412d-893c-438a36ba5a3f

After:

https://github.com/user-attachments/assets/5ce69c04-3188-4484-94f2-58429be333a8

### DIFF
--- a/src/elements/Input/Input.tsx
+++ b/src/elements/Input/Input.tsx
@@ -317,7 +317,7 @@ export const Input = forwardRef<InputRef, InputProps>(
       }
     }, [fontFamily, space, color])
 
-    const inputVariants = useMemo(() => getInputVariants(theme), [])
+    const inputVariants = useMemo(() => getInputVariants(theme), [theme])
 
     useEffect(() => {
       const inputState = getInputState({


gkartalis (approved):
lol thanks for the fix!

MounirDhahri (approved):
This might break things if they switches between dark mode and light mode on a screen that has an input but that's something we don't have currently so we can go with this.

src/elements/Input/Input.tsx
```diff
@@ -317,7 +317,7 @@ export const Input = forwardRef<InputRef, InputProps>(
       }
     }, [fontFamily, space, color])
 
-    const inputVariants = getInputVariants(theme)
+    const inputVariants = useMemo(() => getInputVariants(theme), [])
```
MrSltun, line 320: Maybe we could update the dependencies for the useMemo here?

```suggestion
    const inputVariants = useMemo(() => getInputVariants(theme), [theme])
```

MrSltun (approved):
Thank you Carlos!! I just left a non-blocking suggestion :)

artsyit:
:rocket: PR was released in [`v14.0.35`](https://github.com/artsy/palette-mobile/releases/tag/v14.0.35) :rocket:

PR #332, opened 2025-04-22
chore: fix auto releases from labels

MrSltun:
This PR resolves [] 

### Description

This PR attempts to fix releases from labels

for the sake of testing (and merging #331) I will add the Major label to this PR

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -1,11 +1,8 @@
 {
+  "extends": "@artsy",
   "plugins": [
     "npm",
     "first-time-contributor",
     "released"
-  ],
-  "author": {
-    "name": "Artsy",
-    "email": "it@artsymail.com"
-  }
+  ]
 }


.autorc
```diff
@@ -1,11 +1,8 @@
 {
+  "extends": "@artsy",
```
MrSltun, line 2: This will get the config from [Artsy's shared auto config](https://github.com/artsy/auto-config)

gkartalis: lets seeee

gkartalis (approved):
thanks

MrSltun:
It worked but also failed? 🤷

MrSltun:
It seems that palette mobile was released twice 😅  and I found the issue, it's because we are using an auto-release plugin twice ([here](https://github.com/artsy/palette-mobile/blob/main/.autorc#L4), and [here](https://github.com/artsy/auto-config/blob/main/package.json#L51)), which made the release happen twice and broke the versions, sorry 🙇 

I'll create a new PR to remove the duplicated plugins, and that should fix it

artsyit:
:rocket: PR was released in [`v17.0.0`](https://github.com/artsy/palette-mobile/releases/tag/v17.0.0) :rocket:

PR #337, opened 2025-04-24
chore: support hard coded colours

MounirDhahri:
This PR resolves [] 

### Description

This PR is inspired by this https://github.com/artsy/eigen/pull/11977/

Here, the goal is to allow us to use colours that are not available as tokens as they are instead of `black` (interpreted by `color` being interpreted as `undefined`)

https://github.com/user-attachments/assets/dd0af5e3-8e65-489b-b131-9e37ca233dcb


src/utils/hooks/useTheme.tsx
```diff
@@ -57,5 +57,13 @@ const color =
...
 
+    const color = (theme.colors as { [key: string]: string })[colorName as Color]
+
+    // The color is not defined in the theme, so we return it as it is
+    // This is useful when we would like to force the usage of a color
+    // Regardless of the theme.
+    if (color === undefined) {
+      return colorName
+    }
     return (theme.colors as { [key: string]: string })[colorName as Color]
```
olerichter00, line 68: Or simply

```
    // return the theme color or the string that was passed as a default (e.g., to support `white`).
    return (theme.colors as { [key: string]: string })[colorName as Color] ?? colorName

```

MounirDhahri: sure, if that feels more readable to you. I wanted to have it separately because I felt it deserved its own separate comment

src/utils/hooks/useTheme.tsx
```diff
@@ -57,5 +57,6 @@ const color =
       return undefined
     }
 
-    return (theme.colors as { [key: string]: string })[colorName as Color]
+    // return the theme color or the string that was passed as a default (e.g., to support `white`).
+    return (theme.colors as { [key: string]: string })[colorName as Color] ?? colorName
```
olerichter00, line 61: You are right, separating the returns is a good idea. I was just confused about the last line that didn't use the themeColor stored in the const. What do you think about this version?

```suggestion
    const themeColor = (theme.colors as { [key: string]: string })[colorName as Color]
    
    // return the color that was passed if it's not supported by the theme (e.g., for a theme-independent `white`).
    if(!themeColor) {
      return colorName
    }

    return themeColor
```

olerichter00 (approved):
🥳

src/utils/hooks/useTheme.tsx
MounirDhahri: oups, i missed your comment

artsyit:
:rocket: PR was released in [`v17.2.1`](https://github.com/artsy/palette-mobile/releases/tag/v17.2.1) :rocket:

PR #352, opened 2025-06-04
fix: dark mode issues on Chip

MounirDhahri:
This PR resolves [] 

### Description

This PR fixes an issue with the Chip component breaking when switching from light to dark mode. This was happening because we were using the static configuration, but since we released dark mode, It's not valid anymore.

Screenshot 2025-06-04 at 11 20 59

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/e1501d3e-e233-4281-9469-131400a6d5c8" /> | <video src="https://github.com/user-attachments/assets/7144616d-9138-41f8-9bb1-14f4a93ad82d" /> |

### DIFF
--- a/src/elements/Chip/Chip.tsx
+++ b/src/elements/Chip/Chip.tsx
@@ -1,5 +1,5 @@
-import { useAnimationState, View } from "moti"
-import { FC } from "react"
+import { View } from "moti"
+import { FC, useState } from "react"
 import { useColor, useSpace } from "../../utils/hooks"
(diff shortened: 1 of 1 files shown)


olerichter00 (approved):
Thank you 🙏

MounirDhahri:
you definitely did nothing wrong, I think you could not have built a better API there as I mentioned here. Things changed only the past month after the dark mode release
Screenshot 2025-06-04 at 13 24 06

PR #356, opened 2025-06-18
chore: export image skeleton

MounirDhahri:
### Description

This PR exports `ImageSkeleton` to be used for image loaders.

Related PR: https://github.com/artsy/eigen/pull/12230

PR #357, opened 2025-06-20
build(deps): fix blurhash peerDepencency

gkartalis:
This PR resolves [] 

### Description

Try to fix the issue that started happening after [this PR ](https://github.com/artsy/palette-mobile/pull/356/files)that bumped blurhash in palette-mobile.

We might also need to add:

```
jest.mock("react-native-blurhash", () => {
  const ReactNative = require("react-native")
  return {
    Blurhash: ReactNative.View as any,
  }
})
``` 

in palette but will check it on the automated pr

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   "peerDependencies": {
     "react": "*",
     "react-native": "*",
-    "react-native-blurhash": "^1.1.11",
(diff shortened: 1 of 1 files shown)


MrSltun (approved):

MounirDhahri (approved):
thanks!

PR #364, opened 2025-07-11
chore: use @d11/react-native-fast-image

MounirDhahri:
#future-friday

### Description

This PR adds `@d11/react-native-fast-image` instead of `react-native-fast-image`.

This is for two reasons:
- `react-native-fast-image` hasn't been maintained actively and the last release was 3 years ago. See https://github.com/DylanVann/react-native-fast-image
  - Since images are an integral part of the browsing experience, it woud introduce a liability over the long term
 - `@d11/react-native-fast-image` supports the new architecture
 - `@d11/react-native-fast-image` uses the exact same API

**What else can we consider in the future?**
- `expo-fast-image`. But that's something for another time

**What comes next?**
- I will introduce the same change to `Eigen` and `Folio`

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -10,15 +10,23 @@ PODS:
   - hermes-engine (0.76.9):
     - hermes-engine/Pre-built (= 0.76.9)
   - hermes-engine/Pre-built (0.76.9)
-  - libwebp (1.2.4):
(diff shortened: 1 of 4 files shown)


artsyit:
:rocket: PR was released in [`v17.26.0`](https://github.com/artsy/palette-mobile/releases/tag/v17.26.0) :rocket:

PR #369, opened 2025-08-08
feat: pass scrollviewprops to screen body

gkartalis:
This PR resolves [] 

### Description

Add `scrollViewProps` support to Screen.Body, intended for use in energy.

The reason for this is to be able to use `keyboardShouldPersistTaps="handled"`  so we can save some time when login in and not having to press the button twice to work.


artsyit:
:rocket: PR was released in [`v17.30.0`](https://github.com/artsy/palette-mobile/releases/tag/v17.30.0) :rocket:

PR #372, opened 2025-08-25
feat: migrate to expo

MounirDhahri:
### Description

This PR migrates Palette-mobile to Expo Prebuild.


MounirDhahri:
@MrSltun yes - that's something @araujobarret is looking into and the next step I believe. For the sake of not getting this PR more complicated, I suggest to keep that separate because it's also a big task

MounirDhahri:
@araujobarret related to my comment above - maybe you stack over this branch since we probably won't be merging this as it is

MounirDhahri:
Yess @gkartalis

PR #391, opened 2025-09-30
chore(Revert): "chore: upgrade reanimated to 3.19.1"

gkartalis:
# Description

Reverts artsy/palette-mobile#390

the reason for this is that react native of palette-mobile is not compatible with this reanimated version (see compatibility [table here](https://docs.swmansion.com/react-native-reanimated/docs/guides/compatibility/))


artsyit:
:rocket: PR was released in [`v19.10.0`](https://github.com/artsy/palette-mobile/releases/tag/v19.10.0) :rocket:

PR #394, opened 2025-10-01
chore: upgrade storybook v8

MounirDhahri:
This PR resolves [PHIRE-2347] 

### Description

This PR upgrades storybook to v8.

This upgrade was pretty straighforward, it was almost entirely automated. Great DX 👏🏽 

Steps taken:
- Follow changes here https://github.com/chromaui/intro-storybook-react-native-template/pull/5

### Note 
- The new version uses react-native-gesture-handler and bottom-sheet by default as peer dependencies. This is maybe fine here but imo, I wish they allowed devs to choose at least. Both packages are dev packages

https://github.com/user-attachments/assets/9b638ca0-d849-4c21-a7ed-0d9931ba0b04

[PHIRE-2347]: https://artsyproduct.atlassian.net/browse/PHIRE-2347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -40,9 +40,12 @@
     "@styled-system/core": "^5.1.2",
     "@styled-system/theme-get": "^5.1.2",
     "events": "^3.3.0",
+    "expo": "~52.0.47",
```
MounirDhahri, line 43: @gkartalis there is something that is triggering this addition after install - I actually didn't add it here as well

package.json
```diff
@@ -143,7 +145,6 @@
   "files": [
     "dist"
   ],
-  "main": "dist/index.js",
```
gkartalis, line 146: blocking: this **should remain** in the package.json file

package.json
```diff
@@ -124,9 +128,10 @@
     "react-dom": "18.3.1",
     "react-native": "0.76.9",
     "react-native-device-info": "14.0.4",
+    "react-native-gesture-handler": "~2.22.1",
     "react-native-haptic-feedback": "1.14.0",
     "react-native-linear-gradient": "2.6.2",
-    "react-native-reanimated": "3.16.7",
+    "react-native-reanimated": "3.17.5",
```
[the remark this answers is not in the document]
gkartalis: this is actually not the same version as eigen as of yesterday https://github.com/artsy/eigen/blob/main/package.json#L183

package.json
```diff
@@ -76,9 +76,10 @@
     "@babel/preset-react": "7.18.6",
     "@babel/preset-typescript": "7.18.6",
     "@babel/runtime": "^7.25.0",
+    "@gorhom/bottom-sheet": "^4.6.4",
```
gkartalis, line 79: not a huge fan of adding extra dependencies for a dependency but since it is a dev dep not a big deal I guess

package.json
MounirDhahri: lol this is so annoying 😄  I keep adding it and removing it and it slips back in

package.json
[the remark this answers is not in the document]
MounirDhahri: cool - this was the version that we were at in Eigen so it seemed safe to do it but good point

artsyit:
:rocket: PR was released in [`v19.15.0`](https://github.com/artsy/palette-mobile/releases/tag/v19.15.0) :rocket:

package.json
MounirDhahri: OMG this slipped again - I hate it

PR #392, opened 2025-09-30
build(deps): bump expo and react native

gkartalis:
This PR resolves [PHIRE-2349]

### Description

This PR bumps Expo to SDK53 and RN to 79

| Android | Header |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/ee93096f-73f4-422a-bcd5-b1be668edad9" /> | <video src="https://github.com/user-attachments/assets/0f6dcadc-72a1-446f-929e-39237c7f71b9" /> | 

[PHIRE-2349]: https://artsyproduct.atlassian.net/browse/PHIRE-2349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<details>
  <summary>📦 Published PR as canary version: <code>19.13.0--canary.392.4627.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @artsy/palette-mobile@19.13.0--canary.392.4627.0
  # or 
  yarn add @artsy/palette-mobile@19.13.0--canary.392.4627.0
  ```
</details>

PR #396, opened 2025-10-02
fix: remove jotai and refactor to use useState in order to fix palette-mobile

gkartalis:
This PR resolves [] 

### Description

Realised that palette-mobile wasn't running locally after my [expo bump PR](https://github.com/artsy/palette-mobile/pull/392) and the error was looking something like this:

Screenshot 2025-10-02 at 17 50 30

Located the issue to be in jotai and I just refactored the storybook decorator to use useState instead of jotai and got rid of the dependency since we don't need another state manager :P

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
   "files": [
     "dist"
   ],
+  "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "publishConfig": {
     "access": "public"
--- a/package.json
+++ b/package.json
@@ -114,7 +114,6 @@
     "jest-environment-jsdom": "29.4.2",
     "jest-extended": "3.2.3",
     "jest-watch-typeahead": "2.2.2",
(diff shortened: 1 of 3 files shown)


MrSltun (dismissed):

gkartalis:
ignore the previous message - last commit addressed these issues and we are good to go 📦

artsyit:
:rocket: PR was released in [`v19.14.0`](https://github.com/artsy/palette-mobile/releases/tag/v19.14.0) :rocket:

package.json
```diff
@@ -114,7 +114,6 @@
     "jest-environment-jsdom": "29.4.2",
     "jest-extended": "3.2.3",
     "jest-watch-typeahead": "2.2.2",
-    "jotai": "^2.15.0",
```
brainbicycle, line 117: 👋

PR #477, opened 2026-04-21
chore(ONYX-2058): improve version syncing script

dariakoko:
This PR resolves [ONYX-2058] 

### Description

`auto shipit` only bumps the version in root `package.json` by design - syncing `Example/package.json` and `Example/app.json` at release time isn't possible. This means those files are always out of sync until deploy time, and the previous `check_version_synced` guard would frequently block deploys, requiring developers to manually fix and recommit before they could proceed.  
                                                                                
This removes that guard and replaces it with a `sync-example-version` script that runs automatically as part of `yarn deploy-beta`. It fetches the latest version from `origin/main`, updates all three files, and commits the changes before pushing to the `beta-ios` and `beta-android` branches that trigger the store builds - no manual steps needed.

[ONYX-2058]: https://artsyproduct.atlassian.net/browse/ONYX-2058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MrSltun (dismissed):
Looks good 🌟

MounirDhahri (dismissed):
This looks good to me - Did you see if there is a way with auto to run the script as a post bump hook? that would save us a ton of synching in this case.

dariakoko:
@MounirDhahri I initially added npm `version` lifecycle script in package.json. But I was not sure about it
Here, I committed it so you can see what i mean https://github.com/artsy/palette-mobile/pull/477/changes/3c48a0b8a463ee3a3fc7bd08581a85cea43eeaec

MounirDhahri:
thanks for sharing that - I was mostly referring to [Auto release lifecycle hooks](https://intuit.github.io/auto/docs/plugins/release-lifecycle-hooks#beforeshipit). I am not sure which one would be the right one but I would assume it is possible to do it there. But it's fine as well if that's not possible

dariakoko:
Ah I see, thanks for sharing

I don't think this is suitable for our setup - `beforeShipIt` fires before the new version is known, as far as I understood, so we would have to calculate/get it somehow, I find it tricky

**npm version lifecycle hook** should be pretty okay, but I'm a bit confused with the way we set the version number, so I'm not 100% certain it would work reliably 
  
So, the approach with **deploy-beta script** that fetches main and commits is the most straightforward in my opinion

MounirDhahri:
> I don't think this is suitable for our setup - `beforeShipIt` fires before the new version is known, as far as I understood, so we would have to calculate/get it somehow, I find it tricky. 
I just realized the link I used have a `href` to `beforeShipIt` but I was referring in general to auto lifecycle hooks. [afterVersion](https://intuit.github.io/auto/docs/plugins/release-lifecycle-hooks#afterversion) might be one to consider here.

> npm version lifecycle hook should be pretty okay, but I'm a bit confused with the way we set the version number, so I'm not 100% certain it would work reliably
I haven't investigated this further, but fwiw, the versioning happens based on the label (patch, minor, major) 

> So, the approach with deploy-beta script that fetches main and commits is the most straightforward in my opinion
Sure, happy to keep this as I mentioned earlier. I saw those scripts getting bigger and bigger in Eigen and it's why I thought of the lifecycle as way to simplify things

artsyit:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Messages</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>:tada: - congrats on your new release</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against be8c3ac23192109dc530536fb2e5d13b1c050f1e
</p>

dariakoko:
Her I'm adding `afterVersion` (claude generated code): https://github.com/artsy/palette-mobile/pull/477/changes/be8c3ac23192109dc530536fb2e5d13b1c050f1e
I find it harder to understand, but it could be because I never worked with github plugins before. Happy to merge this version, we can always fall back to the initial approach 
In the end, they work the same - generate an extra commit with the version bump

MounirDhahri:
@dariakoko I will leave it to you to decide on how to proceed with this - both way to me are fine as I mentioned. Thanks a lot for trying this out

artsyit:
:rocket: PR was released in [`v23.12.0`](https://github.com/artsy/palette-mobile/releases/tag/v23.12.0) :rocket:

PR #484, opened 2026-06-03
fix: runtime error + stories for placeholderarray

gkartalis:
This PR resolves [] 

### Description

There was a runtime error in eigen on inputs that had array placeholder and this PR fixes it, see eigen video below for before and after.

Also adds a few stories for array placeholders that we didn't have before

|Before|After|
|---|---|
|<video src="https://github.com/user-attachments/assets/f2c6610a-a502-4183-827c-1efa4048129c" width="400" />|<video src="https://github.com/user-attachments/assets/2684b90b-2e18-4117-a39a-3c30bade0e2d" width="400" />|

### DIFF
--- a/src/elements/Input/Input.tsx
+++ b/src/elements/Input/Input.tsx
@@ -723,7 +723,7 @@ export const Input = forwardRef<InputRef, InputComponentProps>(
         >
           {placeholder.map((placeholderString, index) => (
             <Text
-              key={index}
+              key={`${placeholderString.replaceAll(" ", "-")}-${index}`}
               onLayout={(event) => {
                 placeholderWidths.current[index] = event.nativeEvent.layout.width
               }}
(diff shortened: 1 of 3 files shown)


src/elements/Input/Input.tsx
```diff
@@ -725,6 +723,7 @@ export const Input = forwardRef<InputRef, InputComponentProps>(
         >
           {placeholder.map((placeholderString, index) => (
             <Text
+              key={index}
```
araujobarret: to be on the safe side, we could have:
```tsx
key={`${placeholderString.replaceAll(" ", "-")}-${index}`}
```
just to avoid conflicting keys, usually linters also flag to not use index as keys

araujobarret (dismissed):

artsyit:
:rocket: PR was released in [`v23.14.0`](https://github.com/artsy/palette-mobile/releases/tag/v23.14.0) :rocket:

PR #503, opened 2026-07-30
fix(DI-498): onboarding pill scales with accessibility font size

iskounen:
This PR resolves [DI-498]

### Description

The `onboarding` Pill variant hardcoded a fixed `height: 40px`, so text grown via large accessibility font sizes (or wrapped onto a second line) got clipped instead of the pill growing to fit it.

- Swapped the fixed height for `padding-horizontal`/`padding-vertical` so the pill sizes itself from content instead of a fixed value, scoped to the `onboarding` variant only — no other variant's styles changed.
- Added multiline-aware text alignment: the label stays centered for a single line (unchanged look), but switches to left-aligned once `onTextLayout` reports more than one line, since centered text looks unbalanced once wrapped. Also scoped to `onboarding` only.
- Added tests covering the alignment switch and confirming other variants are unaffected.

Before/after screenshots and the original diagnosis live in the paired POC PR in eigen: https://github.com/artsy/eigen/pull/13862

### Test plan
- [x] `yarn test src/elements/Pill` — 4/4 passing
- [x] `yarn type-check`
- [x] `yarn lint`

### DIFF
--- a/src/elements/Pill/Pill.tests.tsx
+++ b/src/elements/Pill/Pill.tests.tsx
@@ -1,4 +1,5 @@
 import { fireEvent, render, screen } from "@testing-library/react-native"
+import { StyleSheet } from "react-native"
 import { Pill } from "./Pill"
 import { Theme } from "../../Theme"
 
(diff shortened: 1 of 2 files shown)


src/elements/Pill/Pill.tsx
```diff
@@ -80,6 +81,12 @@ export const Pill: React.FC<PillProps> = ({
         <Text
           variant="xs"
           color={color}
+          textAlign={variant === "onboarding" ? (isMultiline ? "left" : "center") : undefined}
+          onTextLayout={
+            variant === "onboarding"
```
mbrowne, line 86: Is there a more generic name for this variant than "onboarding" that might work here? Or just leave as-is if we don't anticipate sharing these styles for any other use cases.

iskounen: I agree that the naming should be more generic. It's a variant name that was added when the (now) old onboarding flow was built, and it would have been better if it had been named "option" or "selection" to match the other variants.

I'm going to leave it as-is for this PR.

artsyit:
:rocket: PR was released in [`v24.4.0`](https://github.com/artsy/palette-mobile/releases/tag/v24.4.0) :rocket:

src/elements/Pill/Pill.tests.tsx
```diff
@@ -30,4 +31,34 @@ describe("Pill", () => {
...
   })
+
+  it("switches the onboarding variant's text alignment from center to left once it wraps to multiple lines", () => {
+    render(
+      <Theme>
+        <Pill variant="onboarding">Some label</Pill>
+      </Theme>
+    )
+
+    expect(StyleSheet.flatten(screen.getByText("Some label").props.style).textAlign).toBe("center")
```
anandaroop, line 42: TIL this kind of assertion.

anandaroop (commented):
Nice!

PR #504, opened 2026-07-31
fix(DI-498): split onboarding pill into new option variant, left-align text

iskounen:
This PR resolves [DI-498]
Assisted-by: Claude: Sonnet-5

### Description

Splits the onboarding-variant `Pill` (shared by the new and old onboarding flows) into `"onboarding"` (reverted to its original fixed-height layout, used by the old flow) and a new `"option"` variant (auto-grows to fit content, always left-aligns text). Also adds an optional `color` prop for callers to override text color when their background doesn't match the variant's themed default.

I introduced this split because the changes in #503 made the pills in the old onboarding flow look comically large, and I was spending a lot of time going back and forth between the two flows in order to reconcile them. This split allows me to (1) focus on the new onboarding flow and (2) create a more generic variant that is not tied to either onboarding flow.

### DIFF
--- a/src/elements/Pill/Pill.stories.tsx
+++ b/src/elements/Pill/Pill.stories.tsx
@@ -121,6 +121,21 @@ export const Onboarding = () => {
   )
 }
 
+export const Option = () => {
+  return (
+    <List contentContainerStyle={{ marginHorizontal: 20, gap: 1 }}>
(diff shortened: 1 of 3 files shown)


artsyit:
:rocket: PR was released in [`v24.5.0`](https://github.com/artsy/palette-mobile/releases/tag/v24.5.0) :rocket:

